PR #4, opened 2023-06-07
Add batch withdraw and natspec

pandadefi:


contracts/StreamManager.vy
```diff
@@ -94,18 +94,33 @@ def __init__(
...
+    """
     assert msg.sender == self.owner
     self.token_is_accepted[token] = True
 
 
 @external
 def remove_token(token: ERC20):
+    """
+    @dev Remove a token from the list of accepted tokens. This can only be
+        called by the owner of the contract.
```
fubuloubu, line 121:
```suggestion
    @dev Remove a token from the list of accepted tokens.
    @notice This can only be called by the owner of the contract.
```

contracts/StreamManager.vy
```diff
@@ -235,14 +310,17 @@ def cancel_stream(
     return funded_amount - amount_locked
 
 
-@external
+@internal
 def withdraw(creator: address, stream_id: uint256) -> uint256:
```
fubuloubu, line 314:
```suggestion
def _withdraw(creator: address, stream_id: uint256) -> uint256:
```

contracts/StreamManager.vy
```diff
@@ -252,3 +330,25 @@ def withdraw(creator: address, stream_id: uint256) -> uint256:
...
+
+@external
+def withdraw_all(creators: address[]):
+    """
+    @dev Withdraw from all streams for a given creator.
+    @param creators The creator of the stream.
+    """
+    for creator in creators:
+        for stream_id in range(self.num_streams[creator]):
+            self.withdraw(creator, stream_id)
```
fubuloubu, line 353:
```suggestion
            self._withdraw(creator, stream_id)
```

contracts/StreamManager.vy
```diff
@@ -117,6 +132,19 @@ def create_stream(
...
 ) -> uint256:
+    """
+    @dev Create a stream targeting the owner of this contract. The stream will
+        be funded with the lesser of the `amount_per_second` * `max_stream_life`
+        or the `funded_amount` of the token. The `start_time` can be in the
+        past, but the stream will not be able to be cancelled until
+        `start_time` + `MIN_STREAM_LIFE` has elapsed. 
+    @param token The token to use for the stream.
+    @param amount_per_second The amount of the token to stream per second.
+    @param reason A reason for the stream. This is optional.
```
fubuloubu, line 143:
```suggestion
    @param reason A reason for the stream.
        Defaults to an empty bytestring.
```

contracts/StreamManager.vy
```diff
@@ -117,6 +132,19 @@ def create_stream(
...
+    @param start_time The time to start the stream. Defaults to the current block
+        timestamp. Default to now.
```
fubuloubu, line 145:
```suggestion
    @param start_time The time to start the stream.
        Defaults to current block timestamp.
```

contracts/StreamManager.vy
```diff
@@ -160,6 +188,14 @@ def create_stream(
 @view
 @internal
 def _amount_unlocked(creator: address, stream_id: uint256) -> uint256:
+    """
+    @dev Get the amount of the stream that is unlocked and available for
+        withdrawal.
+    @param creator The creator of the stream.
+    @param stream_id The id of the stream.
+    @return The amount of the stream that is unlocked and available for
+        withdrawal.
```
fubuloubu, line 197: I usually use 100 chars (god I wish Vyper had a fixer)
```suggestion
    @return The amount of the stream that is unlocked and available for withdrawal.
```

contracts/StreamManager.vy
```diff
@@ -172,12 +208,26 @@ def _amount_unlocked(creator: address, stream_id: uint256) -> uint256:
 @view
 @external
 def amount_unlocked(creator: address, stream_id: uint256) -> uint256:
+    """
+    @dev Get the amount of the stream that is unlocked and available for
+        withdrawal.
+    @param creator The creator of the stream.
+    @param stream_id The id of the stream.
+    @return The amount of the stream that is unlocked and available for
+        withdrawal.
```
fubuloubu, line 217:
```suggestion
    @return The amount of the stream that is unlocked and available for withdrawal.
```

contracts/StreamManager.vy
```diff
@@ -252,3 +330,28 @@ def withdraw(creator: address, stream_id: uint256) -> uint256:
...
+    return self._withdraw(creator, stream_id)
+
+
+@external
+def withdraw_all(creators: DynArray[address, 128], strams: DynArray[uint256, 128]):
+    """
+    @dev Withdraw from all streams for a given creator.
+    @param creators The creator of the stream.
+    """
+    creators_len: uint256 = len(creators)
```
fubuloubu, line 351: honestly don't think this saves much gas vs. using the length directly

contracts/StreamManager.vy
```diff
@@ -252,3 +330,28 @@ def withdraw(creator: address, stream_id: uint256) -> uint256:
...
+    """
+    @dev Withdraw from a stream.
+    @param creator The creator of the stream.
+    @param stream_id The id of the stream.
+    """
+    return self._withdraw(creator, stream_id)
+
+
+@external
+def withdraw_all(creators: DynArray[address, 128], strams: DynArray[uint256, 128]):
```
fubuloubu, line 346: would be nice to create a constant `MAX_BATCH_SIZE` instead of 128

contracts/StreamManager.vy
```diff
@@ -1,4 +1,4 @@
-# @version 0.3.9
+# @version 0.3.7
```
fubuloubu, line 1: Was hoping to keep this 0.3.9, is there something that doesn't work?

contracts/StreamManager.vy
```diff
@@ -237,12 +296,39 @@ def cancel_stream(
...
+    """
+    @dev Withdraw from all streams for a given creator.
+    @param creators The creator of the stream.
+    """
+    assert len(creators) == len(strams), "creators and streams must be the same length"
+
+    for i in range(0, MAX_BATCH_SIZE):
+        if i >= len(creators):
+            break
+        self._withdraw(creators[i], strams[i])
```
fubuloubu, line 318:
```suggestion
        self._withdraw(creators[i], stream_ids[i])
```

contracts/StreamManager.vy
```diff
@@ -237,12 +296,39 @@ def cancel_stream(
...
+
+
+@internal
+def _withdraw(creator: address, stream_id: uint256) -> uint256:
     funded_amount: uint256 = self.streams[creator][stream_id].funded_amount
     withdrawal_amount: uint256 = min(
         self._amount_unlocked(creator, stream_id),
```
fubuloubu, line 325: I noticed that this method already is a `min` of the total funded amount

apepay/__init__.py
```diff
@@ -306,6 +306,12 @@ def create(
             creation_receipt=tx,
         )
 
+    def batch_withdraw(self, creators: List[AddressType], stream_ids: List[int], **txn_kwargs) -> ContractTransactionHandler:
```
fubuloubu, line 309: Might be nicer to accept an input dict of creators to a set of ids, and convert it to the parallel lists

pandadefi: Sure

contracts/StreamManager.vy
```diff
@@ -252,3 +320,43 @@ def withdraw(creator: address, stream_id: uint256) -> uint256:
...
+
+    for i in range(MAX_BATCH_SIZE):
+        if convert(i, uint256) >= len(batches):
+            break
+        creator: address = batches[i].creator
+        stream_ids: DynArray[uint256, MAX_BATCH_SIZE] = batches[i].stream_ids
+        for j in range(MAX_BATCH_SIZE):
+            if convert(j, uint256) >= len(stream_ids):
+                break
+            stream_id: uint256 = stream_ids[j]
```
fubuloubu, line 345:
```suggestion
    for batch in batches:
        for stream_id in batch.stream_ids:
```

And then below just use `batch.creator` instead of `batch`

contracts/StreamManager.vy
```diff
@@ -252,3 +320,43 @@ def withdraw(creator: address, stream_id: uint256) -> uint256:
     log Withdrawn(creator, stream_id, funded_amount == withdrawal_amount, withdrawal_amount)
 
     return withdrawal_amount
+
+
+@external
+def batch_withdraw(
```
fubuloubu, line 326: Might be more expedient to pass in `token: ERC20` to this method, save the memory cost

fubuloubu (commented):
Looking much better

contracts/StreamManager.vy
pandadefi: We still need to load the token on every iteration to asser the stream uses the same token
```
assert token.address == self.streams[creator][stream_id].token.address, "token must be the same for all streams"
```

pandadefi: And the SLOAD is more expensive than writing one more address on-chain?

fubuloubu: Oh, no no, I mean add it as an input argument

fubuloubu:
Note: batch withdraw actually isn't necessary since anyone can call `claim`, so we can use multicall to claim multiple streams at the same time (and doesn't require admin key either)

PR #12, opened 2023-06-09
test: add some more tests and features to py-sdk [SBK-149]

fubuloubu:


sabotagebeats (approved):
LGTM

sabotagebeats (approved):

antazoey:
tests are running for me correctly with this branch, no more weird output corrupted error

tests/test_stream.py
```diff
@@ -47,10 +76,10 @@ def test_create_stream(chain, payer, tokens, stream_manager, extra_args):
     with pytest.raises(apepay_exc.StreamLifeInsufficient):
         stream_manager.create(tokens[0], amount_per_second + 1, sender=payer)
 
-    start_time = chain.pending_timestamp
     stream = stream_manager.create(
         tokens[0], amount_per_second, **extra_args, sender=payer
     )
+    start_time = chain.blocks.head.timestamp
```
antazoey, line 82: you should use `pytest.approx()` perhaps.. this test sometimes fails for me.

fubuloubu: I was actually having issues with how I was doing it before, but this should actually be deterministic because it's the last _mined_ block's timestamp, e.g. the block containing the transaction I just made above. I finally realized this was a much better way to do it vs. tracking `chain.pending_timestamp` prior to making the transaction, since transactions take longer to process in CI

apepay/__init__.py
```diff
@@ -74,11 +82,67 @@ def validators(self) -> List[Validator]:
...
+        else:
+            return item
+
+    def set_validators(
+        self,
+        validators: List[_ValidatorItem],
+        **txn_kwargs,
+    ) -> ReceiptAPI:
+        if len(validators) > 20:
+            raise
```
antazoey, line 106: an empty raise is not valid Python.
There is no exception on the stack.

Try this in a python inpretter:

```python
raise
```

**The only time you can use an empty raise is when there is an exception on the stack - to raise that exception and preserver the exception stack**
like this:

```python
try:
  asfasdf
except Exception:
   print("asdfaasdfasdf")
   raise  # Raise the exception we caught in the except.
```

An empty `raise` by itself in the middle of nowhere is not valid tho, you will get this error:

```
  File "<stdin>", line 1, in <module>
RuntimeError: No active exception to reraise
```

antazoey: Maybe you thought you were writing Vyper?

apepay/__init__.py
```diff
@@ -74,11 +82,67 @@ def validators(self) -> List[Validator]:
...
+            return item.address
+        else:
+            return item
+
+    def set_validators(
+        self,
+        validators: List[_ValidatorItem],
+        **txn_kwargs,
+    ) -> ReceiptAPI:
+        if len(validators) > 20:
```
antazoey, line 105: Shouldn't this be `>=` 20?

The `def validators()` property implies that you can have up to 20 validators.
this check only ensures it is not > 20 at the start
Thus it will attempt to set the 21st but that one won't show up in the `validators()` list

antazoey:
Somehow, the GH action is not using the latest APePay code from this branch when it runs the tests

apepay/__init__.py
antazoey: I resolved this.

apepay/__init__.py
antazoey: I resolved this, let me know I for some reason am fudging up the understanding tho

antazoey:
was this on auto-merge?? I was still debugging stuff oops

PR #15, opened 2023-07-08
test: put back tests [SBK-169]

antazoey:
uncomment out tests and delete todo
and we will continue debugging if this fails

### DIFF
--- a/tests/test_stream.py
+++ b/tests/test_stream.py
@@ -86,6 +86,6 @@ def test_create_stream(chain, payer, tokens, stream_manager, extra_args):
     assert stream.creator == payer
     assert stream.amount_per_second == amount_per_second
     assert stream.reason == extra_args.get("reason", "")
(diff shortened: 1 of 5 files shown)


antazoey:
OK Finally figured this out.
the fix to the real problem occurred here: https://github.com/ApeWorX/ape/pull/1534/files
so to fix it on ApePay, we have to temporarily use ape main

tests/test_stream.py
```diff
@@ -3,94 +3,89 @@
 import pytest
 
 from apepay import exceptions as apepay_exc
-from apepay.utils import TODO_UNDO
 
 
-def test_todo_undo():
```
antazoey, line 9: during some debugging, this test was never meant to be merged, but it did anyway, so just undoing that nonsense.

fubuloubu (approved):

PR #17, opened 2023-07-13
refactor: daemon updates [SBK-173]

fubuloubu:
Got it working on Sepolia

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,8 @@
 .pytest_cache
 __pycache__
 .hypothesis
+
+dist/
+.DS_Store
+
+data.sqlite
--- a/.gitignore
+++ b/.gitignore
@@ -11,5 +11,3 @@ __pycache__
 
 dist/
 .DS_Store
-
-data.sqlite
--- a/apepay/__init__.py
+++ b/apepay/__init__.py
@@ -8,6 +8,7 @@ from ape.contracts.base import ContractInstance, ContractTransactionHandler
 from ape.exceptions import ContractLogicError, DecodingError
 from ape.types import AddressType, ContractLog, HexBytes
 from ape.utils import BaseInterfaceModel, cached_property
+from ethpm_types import ContractType
 from pydantic import validator
 
 from .exceptions import (
@@ -51,6 +52,7 @@ _ValidatorItem = Union[Validator, ContractInstance, str, AddressType]
 
 class StreamManager(BaseInterfaceModel):
     address: AddressType
+    contract_type: Optional[ContractType] = None
 
     @validator("address", pre=True)
     def normalize_address(cls, value: Any) -> AddressType:
@@ -58,9 +60,13 @@ class StreamManager(BaseInterfaceModel):
 
     @property
     def contract(self) -> ContractInstance:
-        # TODO: Fix this
-        return self.project_manager.StreamManager.at(self.address)
-        # return self.chain_manager.contracts.instance_at(self.address)
+        return (
+            self.project_manager.StreamManager.at(self.address)
+            if "StreamManager" in self.project_manager.contracts
+            else self.chain_manager.contracts.instance_at(
+                self.address, contract_type=self.contract_type
+            )
+        )
 
     def __repr__(self) -> str:
         return f"<apepay_sdk.StreamManager address={self.address}>"
@@ -75,17 +81,18 @@ class StreamManager(BaseInterfaceModel):
 
         for idx in range(20):
             try:
-                validators.append(
-                    Validator(
-                        contract=self.project_manager.Validator.at(
-                            self.contract.validators(idx)
-                        )
-                    )
-                )
+                validator_address = self.contract.validators(idx)
 
             except (ContractLogicError, DecodingError):
                 # NOTE: Vyper returns no data if not a valid index
                 break
+
+            validator_contract = (
+                self.project_manager.Validator.at(validator_address)
+                if "Validator" in self.project_manager.contracts
+                else self.chain_manager.contracts.instance_at(validator_address)
+            )
+            validators.append(Validator(contract=validator_contract))
 
         return validators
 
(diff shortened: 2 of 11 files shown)


apepay/__init__.py
```diff
@@ -58,9 +58,44 @@ def normalize_address(cls, value: Any) -> AddressType:
...
+
+        mid_block = (stop_block - start_block) // 2 + start_block
+        # NOTE: biased towards mid_block == start_block
+
+        if start_block == mid_block:
+            for tx in self.chain_manager.blocks[mid_block].transactions:
+                if tx.receipt.contract_address == self.address:
+                    return tx.receipt
+
+            raise  # cannot find receipt
```
antazoey, line 78: you can't do empty raises, that does not work in Python

antazoey (changes requested):
empty raise statements are not valid when there are no exceptions on the stack.

apepay/__init__.py
fubuloubu: It'll work, but yeah it needs to be a better error (also I wanted to upstream this to Ape core as the backup method when you can't get the deployment receipt for a contract)

antazoey: no it will not raise!

antazoey:
```
raise  # Must be done within an except block
raise Exception()  # This is probably what you are zintending to do here
```

apepay/__init__.py
```diff
@@ -58,9 +58,44 @@ def normalize_address(cls, value: Any) -> AddressType:
...
+                start_block=start_block, stop_block=mid_block
+            )
+
+        else:
+            return self._get_contract_receipt(
+                start_block=mid_block + 1, stop_block=stop_block
+            )
+
+    @cached_property
+    def creation_block(self) -> int:
```
fubuloubu, line 93: Hack for https://github.com/ApeWorX/ape/issues/1062 to speed up this:
https://github.com/ApeWorX/ApePay/pull/17/files#diff-0eed34615ea15b0b1032af02a3019ad4398e3850fa1a5797402a70da7f48edb6R275-R277

apepay/__init__.py
```diff
@@ -58,9 +58,44 @@ def normalize_address(cls, value: Any) -> AddressType:
 
     @property
     def contract(self) -> ContractInstance:
-        # TODO: Fix this
-        return self.project_manager.StreamManager.at(self.address)
-        # return self.chain_manager.contracts.instance_at(self.address)
+        try:
+            return self.chain_manager.contracts.instance_at(self.address)
+        except Exception:
```
antazoey, line 63: what types of exceptions do you expect here? wondering if we could mitigate anything upstream?

fubuloubu: Could reduce to the "contract not found" error, only need to fallback on a local network or if we don't have an explorer (although it still won't work unless we somehow bundle the manifest)

apepay/__init__.py
fubuloubu: Can get rid of this anyways with https://github.com/ApeWorX/ape/pull/1548

apepay/__init__.py
fubuloubu: Can remove all of this with https://github.com/ApeWorX/ape/pull/1548

apepay/__init__.py
```diff
@@ -58,9 +58,44 @@ def normalize_address(cls, value: Any) -> AddressType:
 
     @property
     def contract(self) -> ContractInstance:
-        # TODO: Fix this
-        return self.project_manager.StreamManager.at(self.address)
-        # return self.chain_manager.contracts.instance_at(self.address)
+        try:
+            return self.chain_manager.contracts.instance_at(self.address)
```
antazoey, line 62: eventually we will want to include the contracts and reference the project via manifest, right?

apepay/__init__.py
antazoey: - [ ] Release ape 0.6.14 and bump here

apepay/__init__.py
antazoey: - [ ] Release ape 0.6.14 and bump here

apepay/daemon.py
```diff
@@ -1,8 +1,10 @@
 import asyncio
```
antazoey, line 1: maybe we should move this to like an `examples` folder or something and/or replace with utils c-net can use

fubuloubu: yeah I think moving this back to the examples folder, and just having some of the useful bits in a submodule that can be used by the example would be the way to go

.gitignore
```diff
@@ -8,3 +8,8 @@
 .pytest_cache
 __pycache__
 .hypothesis
+
+dist/
+.DS_Store
+
+data.sqlite
```
fubuloubu, line 15: Don't think this is needed?

apepay/__init__.py
```diff
@@ -51,16 +52,21 @@ def validate(self, creator, token, amount_per_second, reason) -> bool:
 
 class StreamManager(BaseInterfaceModel):
     address: AddressType
+    contract_type: Optional[ContractType] = None
```
fubuloubu, line 55: I want to somehow bundle the manifest/contract type with the package so that we always have this as a backup

.gitignore
antazoey: you wanna commit these?

fubuloubu: > you wanna commit these?

There's nothing here that generates this file?

fubuloubu: Not super bothersome, just think it has to do with chaosnet, not this repo

pyproject.toml
```diff
@@ -8,7 +8,7 @@ readme = "README.md"
 
 [tool.poetry.dependencies]
 python = ">=3.8,<4"
-eth-ape = "^0.6.12"
+eth-ape = "^0.6.13"
```
antazoey, line 11: Should we use 0.16.14?

antazoey (approved):

pyproject.toml
fubuloubu:
```suggestion
eth-ape = "^0.6.14"
```

antazoey (approved):

Issue #14, opened 2023-06-12
When installing, `scripts` becomes an install python module [SBK-151]


antazoey:
## What Happened

I installed ApePay and `scripts.` became a python module i can important anywhere in my python environment. Can exclude `scripts` from the package install here? This has highlighted a bug in Core where `ape run` will run the wrong module in this situation... I am running `ape run deploy` but no matter what project I am in, it uses `ApePay`.

antazoey:
idk what is going on

```
(Pdb) inspect.getfile(module)
*** TypeError: <module 'scripts' (namespace)> is a built-in module
(Pdb) module.__path__
_NamespacePath(['/Users/jules/ApeProjects/ApePay/scripts'])
```

why does it think `ApePay.scripts` is a builtin module

fubuloubu:
Yo that's wild lmao

fubuloubu:
@antazoey is this solved with https://github.com/ApeWorX/ape/pull/1478?

antazoey:
> @antazoey is this solved with [ApeWorX/ape#1478](https://github.com/ApeWorX/ape/pull/1478)?

That was my fix in Ape when this happened. But also it no longer seems like it is happening here for maybe different reasons?

antazoey:
Nevermind, when ApePay is installed, if I do `import scripts; scripts.__path__`, it will work and point to ApePay no matter what I am... kinda odd

antazoey:
one potential fix is make a `src` folder, move `apepay` in that, and make that the place to look for python modules instead of the root

antazoey:
I tried both:

```
packages = [{ include = "apepay" }]
```

and

```
exclude = [{ path = "scripts" }]
```

to no avail

fubuloubu:
Can't you do `exclude=["tests/"]`?

antazoey:
> Can't you do `exclude=["tests/"]`?

I spent like over an hour trying everything I possible could like that yesterday and could not get anything to work.
When I have a chance, I will re-try that specific case right there, but I am pretty sure that `include` and `exclude` only include / exclude file glob patterns within the packages it finds. `packages` is the key to control which packages it finds. That is what the doc says too, but I noticed Vyper lang seems to be making this same mistake right now in that you can import Vyper's tests anywhere on your whole OS if you have it installed. I don't see any evidence that `include` and `exclude` work the way they are being used in our repos but I am so new at Poetry that is all a bit confusing for me

antazoey:
[...]

PR #18, opened 2023-07-17
fix: issue causing `scripts/` to be importable outside of this project [SBK-151]

antazoey:
fixes: #14 
Fixes: SBK-151

To test, run:

```python
python -c "import scripts; print(scripts.__path__)"
```
from your home directory or somewhere else that is not this repo

I tried everything I could think of for like an hour to avoid having to make a separate `src/` dir.
I tried using `package`, `include`, and `exclude` to make it so it would not include scripts or tests. It never happened! I don't know why.

But the second i made it look for packages only in the `src` directory, things were fine.


antazoey:
@fubuloubu any other ideas or do we just count this as a behavior of poetry? They don't see it as an issue it seems even though the equivalent fix on setup.py projects works. We either have to change the project structure or not use Poetry or accept that `tests` and `scripts` will always be added to `sys.path`

fubuloubu:
> @fubuloubu any other ideas or do we just count this as a behavior of poetry? They don't see it as an issue it seems even though the equivalent fix on setup.py projects works. We either have to change the project structure or not use Poetry or accept that `tests` and `scripts` will always be added to `sys.path`

I think it's okay (and even somewhat desirable in the case of `scripts/`) since those are unlikely to be real packages:
- https://pypi.org/project/tests/
- https://pypi.org/project/scripts/

antazoey:
> I think it's okay (and even somewhat desirable in the case of `scripts/`) since those are unlikely to be real packages:

Because of the issues it had causes in Ape and the increase of complexity that followed, I am inclined to dislike this behavior.
Nonetheless, we "fixed" it in Ape so it is okay.

antazoey:
closing but be warned.

PR #19, opened 2023-07-19
fix: packages [SBK-179]

antazoey:
**NOTE** this does not actually fix our problem, but I am opening the PR so I can link in a separate issue


antazoey:
related issue: https://github.com/python-poetry/poetry/issues/8213

antazoey:
I guess Poetry does not see this as an issue, even though it works fine when doing the equivalent on `setup.py`.

PR #22, opened 2023-07-28
feat: add JS SDK [SBK-189]

fubuloubu:
NOTE: published https://www.npmjs.com/package/@apeworx/apepay


sdk/js/index.ts
```diff
@@ -0,0 +1,15 @@
+import { getContract, Address, GetContractReturnType, PublicClient, WalletClient, Abi } from "viem";
+import StreamManagerContractType from "../../.build/StreamManager.json";
```
antazoey, line 2: Are we using webpack at all? If we ever do, we can add path resolves so we can do:

```
import StreamManagerContractType from "@build/StreamManager.json";
```

which makes managing relative imports a lot easier.

antazoey (commented):
I know this is a bit early to review much, just checking it out

sdk/js/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "@apeworx/apepay",
+  "version": "0.1.0-alpha.0",
+  "author": "ApeWorX LTD.",
+  "license": "Apache-2.0",
+  "type": "module",
+  "devDependencies": {
+    "typescript": "^5.1.6"
```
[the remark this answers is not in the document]
fubuloubu: So, using workspaces I think the top level package.json's `dependencies` are actually dev dependencies for the entire project (keeping them aligned), and then down here this would be prod dependencies for the package itself

That being said, I think this should be removed because typescript is only a dev dependency

ui/lib/CreateStream.tsx
```diff
@@ -0,0 +1,132 @@
...
+    );
+
+    // NOTE: This function should move away from this component
+    props.registerStream(stream);
+  };
+
+  return (
+    <div id="CreateStream">
+      {/* Display: What is the product they are paying for? With what token? */}
+      {props.cart && props.cart}
```
antazoey, line 88: This line is causing issues for me, unless I comment it out.

fubuloubu: It was a way for me to provide a space for displaying a "cart" (which could be compressed down to form the reason string) but yeah, I'm leaning more towards just providing wagmi-style react hooks with this library and moving the UI elements to the demo app, so people can choose their own components when installing

antazoey:
You need to delete the root `apepay` folder, it is not necessary, and it no longer works with the imports anyway

sdk/js/index.ts
```diff
@@ -134,13 +134,12 @@ export default class StreamManager {
       accountOverride || (this.walletClient?.account?.address ?? "0x0");
     // NOTE: 0x0 shouldn't ever be the value of `account` because of the above error
 
-    const args: Array<number | string | Address> = [token, amountPerSecond];
-
+    const args: Array<number | string | Address | ByteArray> = [token, amountPerSecond];
     if (startTime) {
-      args.push(reason || ""); // NOTE: Needs to make sure to have 4 args
+      args.push(reason ? stringToHex(reason) : ""); // NOTE: Needs to make sure to have 4 args
```
antazoey, line 139: probably needs more work to handle all the cases the python one handles

ui/lib/UpdateStream.tsx
```diff
@@ -0,0 +1,7 @@
+import * as React from "react";
+
+const UpdateStream = () => {
```
antazoey, line 3: it'd be super helpful if this was implemented
Reason: You can't really delete streams. You can cancel them, but they don't go away. It'd be nice to detect an existing stream in the App and have the modal prompt you to update it (if it was canceled) rather than create another one

fubuloubu: The security properties are that a stream that ran out of funding is the same as being cancelled

antazoey (approved):

PR #24, opened 2023-07-31
fix: all remaining fixes for accepting ApePay [SBK-194]

antazoey:


apepay/__init__.py
```diff
@@ -337,6 +339,7 @@ def funding_rate(self) -> Decimal:
         """
         return Decimal(self.amount_per_second) / Decimal(10 ** self.token.decimals())
 
+    @property
     def funding_estimate(self, period: timedelta) -> int:
```
fubuloubu, line 343: I added the arg to help you figure out how much the amount per second would be in days or weeks or whatever period you were interested in

antazoey: oh I see that now, ok will put that back.
But I am going to keep this PR around to contain all the small fixes I need

antazoey: i think it may help to have more method-like name here

antazoey: changed it to `estimate_funding`, so the verb is first

fubuloubu: Nice, just rename the PR

fubuloubu (approved):

PR #31, opened 2023-08-11
feat(contracts): deploy factory via create2 [SBK-211]

fubuloubu:
Had to add method to deploy using https://github.com/pcaversaccio/create2deployer for both the StreamManager blueprint and the factory contract

~~Currently debuging why it doesn't work properly to deploy with~~

### DIFF
--- a/scripts/deploy.py
+++ b/scripts/deploy.py
@@ -24,7 +24,8 @@ def cli():
 @ape_cli_context()
 @click.option("--blueprint", default=None)
 @click.option("--create2", default=None, help="A string tag for the create2 deployment salt")
-def factory(cli_ctx, account, network, blueprint, create2):
+@click.option("--publish", is_flag=True)
+def factory(cli_ctx, account, network, blueprint, create2, publish):
     if create2:
         # NOTE: This is the deployment address listed on the create2 deployer's github:
         # https://github.com/pcaversaccio/create2deployer/tree/main#deployments-create2deployer
@@ -65,7 +66,7 @@ def factory(cli_ctx, account, network, blueprint, create2):
             )
             factory = project.StreamFactory.at(factory_address)
 
-            if click.confirm("Publish"):
+            if publish:
                 cli_ctx.project_manager.track_deployment(factory)
                 cli_ctx.provider.network.publish_contract(factory.address)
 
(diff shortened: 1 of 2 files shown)


fubuloubu:
Deployed `StreamFactory` to 0x92823EB2DB42b8df354EDB5A1FB3668057e2935D

fubuloubu:
Deployed `StreamManager` via `StreamFactory` to 0x3543Faeeddb7bAbCbBB216B3627f9c5E0C39CE41

scripts/deploy.py
```diff
@@ -66,11 +102,31 @@ def manager(account, network, owner, min_stream_life, validators, tokens):
...
+
+        tx = factory.create(list(validators), token_addresses, sender=account)
+        try:
+            manager = project.StreamManager.at(tx.return_value)
+        except ApeException:
+            manager = project.StreamManager.at(factory.deployments(account))
+
+        cli_ctx.logger.success(f"StreamManager deployed to '{manager.address}'.")
+
+        if click.confirm("Publish"):
```
antazoey, line 120: for any click confirm, it would be nice to have flags for skipping.

scripts/deploy.py
```diff
@@ -21,38 +23,72 @@ def cli():
...
-                + len_bytes
-                + b"\x3d\x81\x60\x0a\x3d\x39\xf3"
-                + blueprint_bytecode
+@click.option("--create2", default=None, help="A string tag for the create2 deployment salt")
+def factory(cli_ctx, account, network, blueprint, create2):
+    if create2:
+        # NOTE: This is the deployment address listed on the create2 deployer's github:
+        # https://github.com/pcaversaccio/create2deployer/tree/main#deployments-create2deployer
+        # TODO: Add SDK for create2 deployer
+        create2_deployer = Contract("0x13b0D85CcB8bf860b6b79AF3029fCA081AE9beF2")
```
antazoey, line 32: I wonder if deploying should be part of the SDK too so it is more testable

antazoey (approved):

scripts/deploy.py
fubuloubu: My idea was that the `manager` address would be the transition point between the scripts and the SDKs. They all reference a specific deployment, those contracts weren't build for add-hoc deployments. Could look into it further down the line though, especially using the factory's `create` method properly

Issue #20, opened 2023-07-20
Contracts not included in published package [SBK-181]


antazoey:
There is no actual way yet to reference the contracts needed to use this project from the package.
Only the  python package is there, and it assumes you are running from within the cloned repo (containing the contracts).

Related comment: https://github.com/ApeWorX/ApePay/pull/17/files#r1269582243

PR #35, opened 2023-08-22
feat(py-sdk): add manifest to package [SBK-239]

fubuloubu:
fixes: #20


sdk/py/apepay/manifest.json
```diff
@@ -0,0 +1 @@
+../../../.build/__local__.json
```
antazoey, line 1: i am confused by this file, it is not really JSON?

sdk/py/apepay/__init__.py
```diff
@@ -54,26 +55,47 @@ def validate(self, creator, token, amount_per_second, reason) -> bool:
...
+    _local_contracts: Dict[str, ContractType] = {}
 
     @validator("address", pre=True)
     def normalize_address(cls, value: Any) -> AddressType:
         return cls.conversion_manager.convert(value, AddressType)
 
-    @cached_property
-    def _local_contracts(self) -> Dict[str, ContractType]:
+    def __post_init__(self):
+        # 0. If pre-loaded, default that type
```
antazoey, line 65:
```suggestion
        # 0. If pre-loaded, default to that type
```

antazoey (approved):

sdk/py/apepay/manifest.json
fubuloubu: This is a soft link back to the build file, so that poetry will bundle it

fubuloubu: when you install it, it ends up in `site-packages/.build/`

fubuloubu: I suppose a better solution is to just cp it into the directory when it gets built :thinking:

antazoey (approved):

fubuloubu:
> need to debug how to get it installed correctly, still not doing it right lol

Okay figured this out, file just wasn't available lol

sdk/py/apepay/manifest.json
fubuloubu: It fails now if you don't build first, so I like this approach as it's more fail-safe

fubuloubu:
Added link to #20 because this resolves that issue

antazoey (approved):
seems good to me!!

.github/workflows/test.yaml
```diff
@@ -13,24 +13,39 @@ concurrency:
...
+  py-test:
+    needs: build
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - uses: foundry-rs/foundry-toolchain@v1
+      - uses: ApeWorX/github-action@v2
+      # Ape build artifacts used for JS testing
```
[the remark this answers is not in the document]
fubuloubu: Just a bad comment, sorry

PR #40, opened 2023-09-14
feat(py-sdk): add `Stream.total_time` property [SBK-279]

fubuloubu:


sdk/py/apepay/__init__.py
```diff
@@ -429,6 +429,17 @@ def amount_left(self) -> int:
     def time_left(self) -> timedelta:
         return timedelta(seconds=self.contract.time_left(self.creator, self.stream_id))
 
+    @property
+    def total_time(self) -> timedelta:
+        info = self.info  # NOTE: Avoid calling contract twice
```
antazoey, line 434: the note is exactly why it is no so great to hide contract calls behind a property

antazoey (approved):

PR #51, opened 2023-10-06
feat: display streamstatus in app [SBK-310]

alexisnsns:
Adding the stream status widget in the demo app, both as a pie chart and as a bar chart


ui/lib/StreamStatusBar.tsx
fubuloubu, line 1: might be nice to combine this with the other status element, and have a prop enum control if you want bar or pie layout

ui/app/src/App.tsx
```diff
@@ -3,14 +3,21 @@ import { useState } from "react";
...
+} from "wagmi";
 import config from "./config";
 // NOTE: Do this or else it won't render (or create your own CSS)
 import "rc-slider/assets/index.css";
 import "./styles.css";
+// import { CreateStream, StreamStatus } from "@apeworx/apepay-react";
+// import { StreamStatus } from "@apeworx/apepay-react";
+import CreateStream from "../../../ui/lib/CreateStream";
+import StreamStatus from "../../../ui/lib/StreamStatus";
+import StreamStatusBar from "../../../ui/lib/StreamStatusBar";
```
fubuloubu, line 21: Would really like to fix this so that its not local imports

ui/app/src/App.tsx
```diff
@@ -54,6 +61,9 @@ function App() {
     return Math.random().toString(36).substring(7);
   };
 
+  const { address } = useAccount();
+  const [streamId, setStreamId] = useState<number | null>(null);
```
fubuloubu, line 65: would be cool to use the SDK's `StreamManager.onStreamCreated` function to install the `StreamStatus` items into the list of streams to review on the main demo page (`{demo-app-url}/{StreamManager-address}/all`)

alexisnsns: OK; I'll create an issue and address this in a separate PR

ui/app/src/App.tsx
alexisnsns: Tried a bunch of different things; @mikeshultz says this is for monorepos; with his help I aliased it.

ui/app/src/App.tsx
```diff
@@ -54,6 +59,9 @@ function App() {
     return Math.random().toString(36).substring(7);
   };
 
+  const { address } = useAccount();
+  const [streamId, setStreamId] = useState<number>(1);
```
fubuloubu, line 63: Why not store the whole stream?
```suggestion
  const [stream, setStream] = useState<Stream | null>(null);
```

ui/app/src/App.tsx
```diff
@@ -93,14 +99,50 @@ function App() {
...
+            stream={
+              new Stream(
+                config.streamManagerAddress as `0x${string}`,
+                address as `0x${string}`,
+                // todo: get streamID
+                streamId,
+                usePublicClient(),
+                useWalletClient()?.data as WalletClient
+              )
+            }
```
fubuloubu, line 136: Then this gets easier (just don't display the component while `stream` is null)
```suggestion
            stream={stream}
```

ui/app/src/App.tsx
```diff
@@ -93,14 +99,50 @@ function App() {
       >
         <CreateStream
           streamManagerAddress={config.streamManagerAddress as `0x${string}`}
-          amountPerSecond={100000000000000}
-          registerStream={(s: Stream) => console.log(s)}
+          amountPerSecond={100}
+          registerStream={(s: Stream) => {
+            console.log(s);
+            setStreamId(s.streamId);
+          }}
```
fubuloubu, line 106: This gets easier too
```suggestion
          registerStream={setStream}
```

ui/app/src/App.tsx
```diff
@@ -93,14 +92,31 @@ function App() {
       >
         <CreateStream
```
fubuloubu, line 93: maybe use `stream` is null to switch between `StreamStatus` and `CreateStream` components

---

This could eventually be on a `/{StreamManager-address}/create` sub-page, using a better routing structure like:

- `/{StreamManager-address}` - display all streams for a particular stream manager address (and load new ones as created), add filter params to filter down streams
- `/{StreamManager-address}/{creator-address}` - display all streams for a particular stream manager, by a particular creator (same as above, load new ones as they come in and add filter params)
- `/{StreamManager-address}/create` - Create a new Stream, re-route to detail view once created (using `registerStream` prop)
- `/{StreamManager-address}/{creator-address}/{stream-id}` - display the details of a particular stream

ui/lib/StreamStatus.tsx
```diff
@@ -1,36 +1,63 @@
...
 }
 
-const StreamStatus = (props: StreamStatusProps) => {
-  const [timeLeft, setTimeLeft] = React.useState(1); // using `1` to start w/ 100% left
-  React.useEffect(() => {
-    props.stream.timeLeft().then(setTimeLeft).catch(console.error);
-  }, [timeLeft]);
+const StreamStatus: React.FC<StreamStatusProps> = ({ stream, chartType }) => {
+  const [timeLeft, setTimeLeft] = useState<number>(1);
+  const [totalTime, setTotalTime] = useState<number>(1);
```
fubuloubu, line 12: might be better to use `null` as a guard to not trigger the interval until the data is fetched via RPC to initialize the component

then can display the component in a "loading..." state (using a spinner or something), to tell the user we are getting the information

ui/lib/StreamStatus.tsx
```diff
@@ -1,36 +1,63 @@
...
+    const intervalId = setInterval(() => {
+      setTimeLeft((prevTimeLeft) => Math.max(prevTimeLeft - 1, 0));
+    }, 1000);
+
+    // Clear the interval when the component is unmounted
+    return () => clearInterval(intervalId);
+  }, [stream]);
+
+  console.log("totaltime " + totalTime);
+  console.log("timeleft " + timeLeft);
```
mikeshultz, line 28: question: Do we want these merged?  Same for the `console.log()` call in `App.tsx`

ui/app/src/App.tsx
```diff
@@ -3,17 +3,16 @@ import { useState } from "react";
...
-
-// import { CreateStream, StreamStatus } from "@apeworx/apepay-react";
-import { StreamStatus } from "@apeworx/apepay-react";
-import CreateStream from "../../../ui/lib/CreateStream";
 import config from "./config";
 // NOTE: Do this or else it won't render (or create your own CSS)
 import "rc-slider/assets/index.css";
 import "./styles.css";
+import CreateStream from "lib/CreateStream";
+import StreamStatus from "lib/StreamStatus";
```
fubuloubu, line 11: Would prefer a solution here that mirrors line 3 above and uses the package name  (but allows dynamically updating those packages as long as they are installed in a specific way for development)

ui/lib/StreamStatus.tsx
```diff
@@ -1,36 +1,63 @@
...
+      {chartType === "bar" ? (
+        <PieChart
+          data={[{ value: timeLeft, color: "#111" }]}
+          totalValue={totalTime}
+          lineWidth={20}
+          background="#bfbfbf"
+          rounded
+          animate
+          label={({ dataEntry }) =>
+            `${Math.floor((100 * dataEntry.value) / totalTime)}%`
```
mikeshultz, line 43: question: Isn't this just `percentageLeft`?

mikeshultz (approved):
Couple nit picks, no blockers.  LGTM

ui/lib/StreamStatus.tsx
alexisnsns: yes nice catch

ui/app/src/App.tsx
alexisnsns: linked to issue #53

PR #60, opened 2023-10-16
feat: cancel stream component [SBK-328]

alexisnsns:
Create the 'cancel stream' & 'update stream' component and add them to the demo app

### DIFF
--- a/sdk/js/index.ts
+++ b/sdk/js/index.ts
@@ -229,7 +229,7 @@ export default class StreamManager {
   async update(
     creator: Address,
     streamId: number,
-    amount: number
+    amount: bigint
   ): Promise<string> {
     if (!this.walletClient)
       throw new Error("Error funding stream: wallet client is not set");
--- a/sdk/js/index.ts
+++ b/sdk/js/index.ts
@@ -63,6 +63,7 @@ export class Stream {
       streamManager,
       creator,
       streamId,
+      // amount_per_second can't be changed once the stream has been created
       BigInt(streamInfo.amount_per_second),
       publicClient,
       walletClient
@@ -119,10 +120,12 @@ export class Stream {
       this.walletClient.account.address != this.creator &&
       this.walletClient.account.address != (await this.streamManager.owner())
     )
+    // Both the owner and the creator of the stream can cancel it
       throw new Error(
         "Error cancelling stream: you are neither the creator nor the owner of the stream."
       );
 
+    // pass args depending on each situation; reason is optional
     const args =
       this.walletClient.account.address != this.creator
         ? [this.streamId, reason || "", this.creator]
--- a/sdk/js/index.ts
+++ b/sdk/js/index.ts
@@ -213,7 +213,6 @@ export default class StreamManager {
     })) as boolean;
   }
 
-  // legacy create function
   async create(
     token: Address,
     amountPerSecond: bigint,
@@ -272,98 +271,6 @@ export default class StreamManager {
     );
   }
 
-  // new create function
-  // async create(
-  //   token: Address,
-  //   amountPerSecond: bigint,
-  //   reason?: string,
-  //   startTime?: number,
-  //   accountOverride?: Address
-  // ): Promise<Stream> {
-  //   if (!accountOverride && !this.walletClient?.account)
-  //     throw new Error("No account");
-
-  //   const account =
-  //     accountOverride || (this.walletClient?.account?.address ?? "0x0");
-
-  //   const args: Array<number | string | bigint | Address | ByteArray> = [
-  //     token,
-  //     amountPerSecond,
-  //   ];
-  //   if (startTime) {
-  //     args.push(reason ? stringToHex(reason) : "");
-  //     args.push(startTime);
-  //   } else if (reason) {
-  //     args.push(stringToHex(reason));
-  //   }
-
-  //   // Step 1: Run the transaction
-  //   const hash = await this.walletClient?.writeContract({
-  //     chain: null,
-  //     address: this.address,
-  //     abi: StreamManagerContractType.abi as Abi,
-  //     functionName: "create_stream",
-  //     args,
-  //     account,
-  //   });
-
-  //   console.log("writeContract params:", {
-  //     chain: null,
-  //     address: this.address,
-  //     abi: StreamManagerContractType.abi,
-  //     functionName: "create_stream",
-  //     args,
-  //     account,
-  //   });
-
-  //   if (hash === undefined)
-  //     throw new Error("Error while processing transaction. Hash undefined.");
-
-  //   // Step 2: Wait for transaction to be confirmed
-  //   const receipt = await this.publicClient.waitForTransactionReceipt({
-  //     hash: hash,
-  //   });
-
-  //   if (!receipt || !receipt.logs || receipt.logs.length === 0) {
-  //     throw new Error("Transaction failed or logs are missing");
-  //   }
-
-  //   // Step 3: Use the last log for creating the Stream instance
-  //   const lastLog = receipt.logs[receipt.logs.length - 1];
-
-  //   return Stream.fromEventLog(
-  //     this,
-  //     lastLog,
-  //     this.publicClient,
-  //     this.walletClient
-  //   );
-  // }
-
-  // new onstreamcreated
-  // onStreamCreated(
-  //   handleStream: (stream: Stream) => null,
-  //   creator?: Address
-  // ): void {
-  //   const onLogs = (logs: Log[]) => {
-  //     Promise.all(
-  //       logs.map((log: Log) =>
-  //         Stream.fromEventLog(this, log, this.publicClient, this.walletClient)
-  //           .then(handleStream)
-  //           .catch(console.error)
-  //       )
-  //     );
-  //   };
-
-  //   this.publicClient.watchContractEvent({
-  //     address: this.address,
-  //     abi: StreamManagerContractType.abi as Abi,
-  //     eventName: "StreamCreated",
-  //     args: creator ? { creator } : {},
-  //     onLogs,
-  //   });
-  // }
-
-  // legacy onstreamcreated
   onStreamCreated(
     handleStream: (stream: Stream) => null,
     creator?: Address
--- a/sdk/js/index.ts
+++ b/sdk/js/index.ts
@@ -92,7 +92,7 @@ export class Stream {
   async totalTime(): Promise<number> {
     const streamInfo = await this.streamInfo();
     return (
-      (Number(streamInfo.funded_amount) / streamInfo.amount_per_second) +
+      Number(streamInfo.funded_amount) / streamInfo.amount_per_second +
       Number(streamInfo.last_pull - streamInfo.start_time)
     );
   }
--- a/sdk/js/index.ts
+++ b/sdk/js/index.ts
@@ -245,4 +245,13 @@ export default class StreamManager {
       throw new Error("Error while processing transactions");
     return hash;
   }
+
+  async stream_is_cancelable(creator: Address, streamId: number): Promise<boolean> {
+    return (await this.publicClient.readContract({
+      address: this.address,
+      abi: StreamManagerContractType.abi as Abi,
+      functionName: "stream_is_cancelable",
+      args: [creator, streamId],
+    })) as boolean;
+  }
 }
--- a/sdk/js/index.ts
+++ b/sdk/js/index.ts
@@ -213,6 +213,7 @@ export default class StreamManager {
     })) as boolean;
   }
 
+  // legacy create function
   async create(
     token: Address,
     amountPerSecond: bigint,
@@ -221,7 +222,7 @@ export default class StreamManager {
     accountOverride?: Address
   ): Promise<Stream> {
     if (!accountOverride && !this.walletClient?.account)
-      throw new Error("No account");
+      throw new Error("Error on create: no account");
 
     const account =
       accountOverride || (this.walletClient?.account?.address ?? "0x0");
@@ -258,7 +259,7 @@ export default class StreamManager {
     });
 
     if (hash === undefined)
-      throw new Error("Error while processing trasactions");
+      throw new Error("Error while processing transaction; hash undefined");
 
     return new Stream(
       this,
@@ -271,6 +272,98 @@ export default class StreamManager {
     );
   }
 
+  // new create function
+  // async create(
+  //   token: Address,
+  //   amountPerSecond: bigint,
+  //   reason?: string,
+  //   startTime?: number,
+  //   accountOverride?: Address
+  // ): Promise<Stream> {
+  //   if (!accountOverride && !this.walletClient?.account)
+  //     throw new Error("No account");
+
+  //   const account =
+  //     accountOverride || (this.walletClient?.account?.address ?? "0x0");
+
+  //   const args: Array<number | string | bigint | Address | ByteArray> = [
+  //     token,
+  //     amountPerSecond,
+  //   ];
+  //   if (startTime) {
+  //     args.push(reason ? stringToHex(reason) : "");
+  //     args.push(startTime);
+  //   } else if (reason) {
+  //     args.push(stringToHex(reason));
+  //   }
+
+  //   // Step 1: Run the transaction
+  //   const hash = await this.walletClient?.writeContract({
+  //     chain: null,
+  //     address: this.address,
+  //     abi: StreamManagerContractType.abi as Abi,
+  //     functionName: "create_stream",
+  //     args,
+  //     account,
+  //   });
+
+  //   console.log("writeContract params:", {
+  //     chain: null,
+  //     address: this.address,
+  //     abi: StreamManagerContractType.abi,
+  //     functionName: "create_stream",
+  //     args,
+  //     account,
+  //   });
+
+  //   if (hash === undefined)
+  //     throw new Error("Error while processing transaction. Hash undefined.");
+
+  //   // Step 2: Wait for transaction to be confirmed
+  //   const receipt = await this.publicClient.waitForTransactionReceipt({
+  //     hash: hash,
+  //   });
+
+  //   if (!receipt || !receipt.logs || receipt.logs.length === 0) {
+  //     throw new Error("Transaction failed or logs are missing");
+  //   }
+
+  //   // Step 3: Use the last log for creating the Stream instance
+  //   const lastLog = receipt.logs[receipt.logs.length - 1];
+
+  //   return Stream.fromEventLog(
+  //     this,
+  //     lastLog,
+  //     this.publicClient,
+  //     this.walletClient
+  //   );
+  // }
+
+  // new onstreamcreated
+  // onStreamCreated(
+  //   handleStream: (stream: Stream) => null,
+  //   creator?: Address
+  // ): void {
+  //   const onLogs = (logs: Log[]) => {
+  //     Promise.all(
+  //       logs.map((log: Log) =>
+  //         Stream.fromEventLog(this, log, this.publicClient, this.walletClient)
+  //           .then(handleStream)
+  //           .catch(console.error)
+  //       )
+  //     );
+  //   };
+
+  //   this.publicClient.watchContractEvent({
+  //     address: this.address,
+  //     abi: StreamManagerContractType.abi as Abi,
+  //     eventName: "StreamCreated",
+  //     args: creator ? { creator } : {},
+  //     onLogs,
+  //   });
+  // }
+
+  // legacy onstreamcreated
   onStreamCreated(
     handleStream: (stream: Stream) => null,
     creator?: Address
@@ -292,9 +385,6 @@ export default class StreamManager {
         )
         .forEach(handleStream);
     };
-
-    logs.map((log) => Stream.fromEventLog(...).then(handleStream).catch(console.error))
-
     this.publicClient.watchContractEvent({
       address: this.address,
       abi: StreamManagerContractType.abi as Abi,
--- a/sdk/js/index.ts
+++ b/sdk/js/index.ts
@@ -32,7 +32,7 @@ export class Stream {
     creator: Address,
     streamId: number,
     publicClient: PublicClient,
-    walletClient?: WalletClient,
+    walletClient?: WalletClient
   ) {
     this.address = address;
     this.creator = creator;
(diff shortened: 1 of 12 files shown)


sdk/js/index.ts
```diff
@@ -205,4 +205,24 @@ export default class StreamManager {
...
+      chain: null,
+      address: this.address,
+      abi: StreamManagerContractType.abi as Abi,
+      functionName: "cancel_stream",
+      args: [streamId, reason, creator],
+      account: creator,
+    });
+    if (result === undefined)
+      throw new Error("Error while processing transactions");
+    return BigInt(result);
```
mikeshultz, line 226: issue: I believe `writeContract()` returns a tx hash here, not an int.  Unless we're using BigInt instead of strings for hashes, this probably isn't what we want.  The call for `create()` is an example:

https://github.com/ApeWorX/ApePay/blob/8af998c05ca5398ab8fd3c66419cc499171aa730/sdk/js/index.ts#L157-L167

ui/app/src/App.tsx
```diff
@@ -55,6 +60,28 @@ function App() {
 
   const [stream, setStream] = useState<Stream | null>(null);
 
+  // Get streamManager to pass it as props to cancelstream
+  const sm = new StreamManager(
+    config.streamManagerAddress as `0x${string}`,
```
mikeshultz, line 66: question: Does `0x${string}` work?  Don't think I've seen this one before.  If you know what it's called, lmk.

ui/lib/CancelStream.tsx
```diff
@@ -1,7 +1,48 @@
-import * as React from "react";
+import React, { useState } from "react";
+import StreamManager from "../../sdk/js/index";
```
mikeshultz, line 2: suggestion(non-blocking): Could use the alias here?

sdk/js/index.ts
```diff
@@ -205,4 +205,44 @@ export default class StreamManager {
...
+      chain: null,
+      address: this.address,
+      abi: StreamManagerContractType.abi as Abi,
+      functionName: "add_funds",
+      args: [creator, streamId, amount],
+      account: creator,
+    });
+    if (timeLeft === undefined)
+      throw new Error("Error while processing transactions");
+    return BigInt(timeLeft);
```
mikeshultz, line 246: issue: Same issue with return var as `create()`

sdk/js/index.ts
```diff
@@ -205,4 +205,44 @@ export default class StreamManager {
...
+  }
+
+  async update(
+    creator: Address,
+    streamId: number,
+    amount: number
+  ): Promise<bigint> {
+    if (!this.walletClient)
+      throw new Error("Error funding stream: wallet client is not set");
+    const timeLeft = await this.walletClient.writeContract({
```
mikeshultz, line 236: question: Should we be handling token approvals (check & set) in the SDK?  This will fail if the proper approval hasn't been set.  Same concern with `create()` which might be out of scope of this PR.

ui/app/src/App.tsx
```diff
@@ -122,6 +150,25 @@ function App() {
...
+        <div>
+          <CancelStream
+            streamId={stream.streamId}
+            reason={reason}
+            creator={stream.creator}
+            sm={sm}
+          />
+        </div>
+      )}
+      {stream && (
```
mikeshultz, line 163: suggestion(non-blocking): I wonder if this should be more intelligent.  Checking against `max_stream_life` to make sure an update would succeed.  This check is what I'm concerned about:

https://github.com/ApeWorX/ApePay/blob/61ad4b17996723b5a806aa21400a33b32c8d28af/contracts/StreamManager.vy#L201-L204

Maybe something to deal with later.

ui/lib/UpdateStream.tsx
```diff
@@ -1,7 +1,62 @@
...
+  return (
+    <div className="update-stream-container">
+      <div className="update-stream-title">
+        {" "}
+        Enter the amount you want to add to the stream:{" "}
+      </div>
+      <input
+        type="number"
+        value={amount}
+        onChange={(e) => setAmount(Number(e.target.value))}
```
mikeshultz, line 43: issue: `amount` here is likely something a human is going to type (e.g. "1.5" to mean 1.5 USDC), but the contract isn't going to read them that way because of the false-decimal handling.  We should be handling human-to-real conversions here, so if someone inputs 22.50 USDC, we would converting it to `22500000` as a BigInt for a 6-decimal token (decimals should be read from the contract or set in config, depending) before sending it to the contract.  IMO, we should store everything as bigints and only render for humans.

Veim's [`parseEther()`](https://viem.sh/docs/utilities/parseEther.html) and [`praseUnits()`](https://viem.sh/docs/utilities/parseUnits.html) and their equivalent format* functions are probably your friends here.

Let me know if you'd like more context on this.

mikeshultz (commented):
Did a code-pass.  I haven't reviewed UI interaction or anything yet.  Let me know if you have any questions.

ui/app/src/App.tsx
alexisnsns: Screenshot 2023-10-17 at 09 25 51

This is an alias from Viem but you're right; I'll use Address directly 👍

mikeshultz: Syntax just through me off.  Looked it up, it's [template literal types](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html).  Neat.

sdk/js/index.ts
mikeshultz: I think the answer is yes, but not in the SDK (which appears to just be a thin contract wrapper).  Here's how it's done in the `CreateStream` component:

https://github.com/ApeWorX/ApePay/blob/61ad4b17996723b5a806aa21400a33b32c8d28af/ui/lib/CreateStream.tsx#L396-L403

Without taking this step to set the ERC20 allowance (if necessary), the call to `add_funds()` will fail.  Let me know if you have any questions.

ui/lib/UpdateStream.tsx
```diff
@@ -1,7 +1,64 @@
...
+  const handleUpdate = async () => {
+    try {
+      const contractAmount = parseUnits(amount, props.tokenDecimals);
+
+      const result = await props.sm.update(
+        props.creator,
+        props.streamId,
+        contractAmount
+      );
+      setResult(`Stream updated. ${result}`);
```
mikeshultz, line 27: issue: Also, there's none of the same transaction handling in this component that `CreateStream` has.

ui/lib/CancelStream.tsx
```diff
@@ -1,7 +1,48 @@
...
+  // console.log(props.sm);
+
+  const handleCancel = async () => {
+    try {
+      const result = await props.sm.cancel(
+        props.streamId,
+        hexReason,
+        props.creator
+      );
+      setResult(`Stream cancelled. Transaction Hash: ${result}`);
```
mikeshultz, line 28: issue: The issue with that tx talked about in tg is that a stream cannot be cancelled until `MIN_STREAM_LIFE` has been exceeded for the stream.

https://github.com/ApeWorX/ApePay/blob/61ad4b17996723b5a806aa21400a33b32c8d28af/contracts/StreamManager.vy#L31

It's probably a good idea to check `stream_is_cancelable` before allowing the user to waste their money here.

https://github.com/ApeWorX/ApePay/blob/61ad4b17996723b5a806aa21400a33b32c8d28af/contracts/StreamManager.vy#L210-L213

sdk/js/index.ts
fubuloubu: The add value to stream functionality should be upstream to the SDK

ui/lib/CancelStream.tsx
fubuloubu: > It's probably a good idea to check `stream_is_cancelable` before allowing the user to waste their money here.

This could probably raise an error in the SDK function `Stream.addFunds(amount)` (mimicking [python SDK](https://github.com/ApeWorX/ApePay/blob/main/sdk/py/apepay/__init__.py#L448) call to [`StreamManager.add_funds`](https://github.com/ApeWorX/ApePay/blob/main/contracts/StreamManager.vy#L195))

ui/lib/UpdateStream.tsx
```diff
@@ -1,7 +1,206 @@
...
+        {isError && <p>You did not confirm the transaction.</p>}
+      </div>
+    );
+  };
+
+  // Stream update function (for step 2)
+  const handleUpdate = async () => {
+    try {
+      setButtonDisabled(true);
+      const result = await props.sm.update(
```
fubuloubu, line 148: `.update` should be under `Stream`

ui/lib/UpdateStream.tsx
```diff
@@ -83,8 +86,12 @@ const UpdateStream: React.FC<UpdateStreamProps> = (props) => {
   const Step1 = () => {
     return (
       <div className="stream-container">
-        {/* TODO: add dynamic unit of time; now in seconds */}
-        <div>Remaining time in the stream: {timeLeft} seconds.</div>
+        {timeLeft !== null ? (
+          <div>Remaining time in the stream: {formatTime(timeLeft)}.</div>
+        ) : (
+          <div>Fetching remaining time...</div>
+        )}
```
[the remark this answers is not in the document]
fubuloubu: We talked about having stream status display a countdown clock instead of percentage (which is better for the pie chart)

ui/lib/UpdateStream.tsx
```diff
@@ -1,7 +1,224 @@
...
+  useAccount,
+  useBalance,
+} from "wagmi";
+import Slider from "rc-slider";
 
-const UpdateStream = () => {
-  return <div />;
+interface UpdateStreamProps {
+  stream: Stream;
+  streamDailyCost: number;
```
mikeshultz, line 13: question: Similar to my other question above.  Is `number` appropriate here?  Should this be `BigInt` if we're dealing with indivisible amounts?

mikeshultz (approved):
Last commentary is non-blocking but should be something to think about and maybe clarify going forward.  

PR LGTM, thanks for addressing my feedback.

ui/app/src/App.tsx
```diff
@@ -154,22 +159,25 @@ function App() {
...
-            creator={stream.creator}
-            sm={sm}
-          />
-        </div>
+        <>
+          <div>
+            <CancelStream
+              stream={stream}
+              reason={reason}
+              sm={sm}
```
fubuloubu, line 166: `stream` should have access to all this info

ui/app/src/App.tsx
```diff
@@ -122,6 +157,29 @@ function App() {
...
+          {cancelResult && <p>{cancelResult}</p>}
+        </>
+      )}
+      {stream && (
+        <div>
+          <UpdateStream
+            stream={stream}
+            sm={sm}
+            token={selectedToken}
+            streamDailyCost={100 * 86400}
```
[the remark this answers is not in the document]
fubuloubu: This is in the Demo app, we can hardcode stuff here for now and open it up over time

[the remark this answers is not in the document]
fubuloubu: For example, the demo app can have a bunch of configurable amounts to change how the component displays

ui/app/src/App.tsx
alexisnsns: Spoke with @fubuloubu ; I will create an issue; max stream life isn't implemented yet so I'll remove it from this PR

sdk/js/index.ts
```diff
@@ -11,7 +11,7 @@ import StreamManagerContractType from "./.build/StreamManager.json";
 
 export interface StreamInfo {
   token: Address;
-  amount_per_second: bigint;
+  amount_per_second: number;
```
fubuloubu, line 14: I would keep this bigint

sdk/js/index.ts
```diff
@@ -20,78 +20,173 @@ export interface StreamInfo {
...
   async totalTime(): Promise<number> {
     const streamInfo = await this.streamInfo();
     return (
-      Number(streamInfo.funded_amount / streamInfo.amount_per_second) +
+      (Number(streamInfo.funded_amount) / streamInfo.amount_per_second) +
       Number(streamInfo.last_pull - streamInfo.start_time)
     );
   }
+
+  async update(amount: bigint): Promise<string> {
```
fubuloubu, line 100: I think `addTime` is a better name

ui/lib/CancelStream.tsx
```diff
@@ -1,7 +1,86 @@
...
+import React, { useState, useEffect } from "react";
+import { Stream } from "../../sdk/js/index";
+import { formatTime } from "./utils";
 
-const CancelStream = () => {
-  return <div />;
-};
+interface CancelStreamProps {
+  stream: Stream;
+  onComplete: (result: string) => void;
```
fubuloubu, line 7: I don't know if this function should pas through the txn hash

ui/lib/CancelStream.tsx
```diff
@@ -1,7 +1,86 @@
...
+}
+
+const CancelStream: React.FC<CancelStreamProps> = (props) => {
+  const [result, setResult] = useState<string | null>(null);
```
fubuloubu, line 11: Probably just an in progress and an error state var is needed

ui/lib/CancelStream.tsx
```diff
@@ -1,7 +1,86 @@
...
+  // Allow user to cancel stream only if he didnt already click on cancel  AND if the stream is cancellable
+  const [isButtonEnabled, setButtonEnabled] = useState(false);
+  // Get the minimum stream life, before which a stream cannot be Canceled
+  const minStreamLife = props.stream.streamManager.MIN_STREAM_LIFE;
+
+  // Check if the stream is cancellable and set the button state accordingly.
+  const checkStreamCancelable = async () => {
+    try {
+      const isCancelable = await props.stream.isCancelable();
```
fubuloubu, line 20: I wonder if we can make this a sync function now that the min stream life is available

ui/lib/CancelStream.tsx
```diff
@@ -1,7 +1,116 @@
...
+import React, { useState, useEffect } from "react";
+import { Stream } from "../../sdk/js/index";
+import { formatTime } from "./utils";
 
-const CancelStream = () => {
-  return <div />;
-};
+interface CancelStreamProps {
+  stream: Stream;
+  onComplete: (error: string | boolean) => void;
```
fubuloubu, line 7: Would this work?
```suggestion
  onComplete: (error: Error | null) => void;
```

Also. Can the component not display its own error?

ui/lib/CancelStream.tsx
```diff
@@ -1,7 +1,116 @@
...
+  }, [startTime]);
+
+  // Fetch current time every second
+  useEffect(() => {
+    const interval = setInterval(() => {
+      setCurrentTime(Date.now() / 1000);
+    }, 1000);
+
+    return () => clearInterval(interval);
+  }, []);
```
fubuloubu, line 85: This seems like a good candidate to move to a utils module e.g. `const currentTime = useCurrentTime()`

ui/lib/CancelStream.tsx
```diff
@@ -1,7 +1,116 @@
...
+  const timeBeforeCancellability = startTime + minStreamLife - currentTime;
+
+  return (
+    <div className="stream-container">
+      <div className="cancel-stream-label">
+        {minStreamLife === null ? (
+          <div>Fetching stream minimum life...</div>
+        ) : !isButtonEnabled && !inProgress && startTime !== 0 ? (
+          <div>
+            Deployment cannot be cancelled yet: its minimum life is
```
fubuloubu, line 97: Might want to eventually add an info popup to clue people into why this is the case

ui/lib/UpdateStream.tsx
```diff
@@ -1,7 +1,202 @@
...
-const UpdateStream = () => {
-  return <div />;
+interface UpdateStreamProps {
+  stream: Stream;
+  onComplete: (error: string | boolean) => void;
+}
+
+const UpdateStream: React.FC<UpdateStreamProps> = (props) => {
+  // Get the token address of the stream
+  const [streamToken, setStreamToken] = useState<`0x${string}` | null>(null);
```
fubuloubu, line 18: With the update I mentioned in the SDK, this is no longer required

ui/lib/CancelStream.tsx
alexisnsns: No we can't have it as null as it would make it impossible to check in the parent component (cf. closing the modal)

alexisnsns: Yes the component (update and cancel) can display their own error

alexisnsns: Edited so that the component displays its own error and we pass True to the parent component when stream is effectively being cancelled/updated

fubuloubu: why is the boolean needed? it's called `onComplete` so I'm assuming the context is that it completed (with or without an error)

alexisnsns: The boolean is needed to set a state in the parent component; so that we'll have something like displayModal={onComplete}

alexisnsns: oh ok I see what you meant

alexisnsns: updated

sdk/js/index.ts
```diff
@@ -20,78 +20,181 @@ export interface StreamInfo {
...
+    if (
+      this.walletClient.account.address != this.creator &&
+      this.walletClient.account.address != (await this.streamManager.owner())
+    )
+      // Both the owner and the creator of the stream can cancel it
+      throw new Error(
+        "Error cancelling stream: you are neither the creator nor the owner of the stream."
+      );
+
+    // pass args depending on each situation; reason is optional
```
fubuloubu, line 133:
```suggestion
    // pass args depending on each situation; reason is optional but must be provided
    // if not using the caller as the creator arg
```

fubuloubu (approved):
looking good

ui/lib/CancelStream.tsx
fubuloubu: but like the whole useEffect setup

sdk/js/index.ts
```diff
@@ -167,32 +260,34 @@ export default class StreamManager {
       throw new Error("Error while processing trasactions");
```
fubuloubu, line 261: can do `await publicClient.waitForTransactionReceipt(hash)` to wait for transaction to fully process

sdk/js/index.ts
```diff
@@ -167,32 +260,34 @@ export default class StreamManager {
...
 
     return new Stream(
```
fubuloubu, line 262: if we await for `transaction` to mine, we can return `await Stream.fromEventLog` using `transaction.logs[-1]` (since it should be the last log emitted) and avoid the RPC call to get `streamId` above while adding better devex around handling failures of this transaction type

sdk/js/index.ts
```diff
@@ -167,32 +260,34 @@ export default class StreamManager {
...
     handleStream: (stream: Stream) => null,
-    creator?: Address,
+    creator?: Address
   ): void {
     const onLogs = (logs: Log[]) => {
       logs
         .map(
           // Log is StreamCreated
           (log: Log) =>
             new Stream(
```
fubuloubu, line 282: this should probably use `Stream.fromEventLog` instead

maybe as like `logs.map((log) => Stream.fromEventLog(...).then(handleStream).catch(console.error))`

ui/lib/utils.tsx
```diff
@@ -0,0 +1,42 @@
...
+    const remainderHours = Math.floor((seconds % 86400) / 3600);
+    return ` ${days} day${days !== 1 ? "s" : ""} and ${remainderHours} hour${
+      remainderHours !== 1 ? "s" : ""
+    }`;
+  }
+}
+
+// Fetch current time in seconds
+export const useCurrentTime = () => {
+  const [currentTime, setCurrentTime] = useState<number>(0);
```
fubuloubu, line 31:
```suggestion
  const [currentTime, setCurrentTime] = useState<number>(Date.now() / 1000);
```

ui/lib/UpdateStream.tsx
```diff
@@ -1,7 +1,184 @@
...
+  onComplete: () => void;
+}
+
+const UpdateStream: React.FC<UpdateStreamProps> = (props) => {
+  // Disable 'update stream' button after a user clicked on it
+  const [isButtonDisabled, setButtonDisabled] = useState(false);
+  // Let users select the number of days they want to fund the stream
+  const [selectedTime, setSelectedTime] = useState(1);
+  // Manage error handling
+  const [Error, setError] = useState<string | null>(null);
```
fubuloubu, line 22: question: isn't `Error` a built-in type?

suggestion: lowercase it

fubuloubu (approved):

sdk/js/index.ts
fubuloubu: we are moving this to another PR

sdk/js/index.ts
fubuloubu: we are moving this to another PR

sdk/js/index.ts
fubuloubu: we are moving this to another PR

sdk/js/index.ts
```diff
@@ -20,78 +20,173 @@ export interface StreamInfo {
...
+      (streamInfo.last_pull - streamInfo.start_time)
     );
   }
+
+  async addTime(amount: bigint): Promise<string> {
+    if (!this.walletClient || !this.walletClient.account)
+      throw new Error("Error funding stream: no wallet client set");
+
+    return this.walletClient.writeContract({
+      chain: null,
```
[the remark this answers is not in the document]
fubuloubu: move this to an issue

sdk/js/index.ts
```diff
@@ -20,78 +20,182 @@ export interface StreamInfo {
...
+  }
+
+  async isCancelable(): Promise<boolean> {
+    return (await this.publicClient.readContract({
+      address: this.streamManager.address,
+      abi: StreamManagerContractType.abi as Abi,
+      functionName: "stream_is_cancelable",
+      args: [this.creator, this.streamId],
+    })) as boolean;
+  }
```
fubuloubu, line 159: can actually make this sync if `StreamInfo.start_time` is added to the properties on the constructor of `Stream`
```suggestion
  isCancelable(): boolean {
    return this.streamManager.MIN_STREAM_LIFE < (Date.now() / 1000 - this.startTime);
  }
```

fubuloubu: moved to #62

PR #67, opened 2023-10-27
fix: time_left max C int issue [SBK-338]

### LINKED ISSUE #65: bug: OverflowError: Python int too large to convert to C int
Seen just now on sepolia.  Will see about instrumenting more tomorrow:

```
Traceback (most recent call last):
  File "/app/.venv/bin/silverback", line 8, in <module>
    sys.exit(cli())
  File "/app/.venv/lib/python3.10/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/app/.venv/lib/python3.10/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/app/.venv/lib/python3.10/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/app/.venv/lib/python3.10/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/app/.venv/lib/python3.10/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/app/.venv/lib/python3.10/site-packages/click/decorators.py", line 92, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "/app/.venv/lib/python3.10/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/app/.venv/lib/python3.10/site-packages/silverback/_cli.py", line 59, in run
    asyncio.run(runner.run())
  File "/usr/local/lib/python3.10/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/usr/local/lib/python3.10/asyncio/base_events.py", line 649, in run_until_complete
    return future.result()
  File "/app/.venv/lib/python3.10/site-packages/silverback/runner.py", line 58, in run
    await self.app.broker.startup()
  File "/app/.venv/lib/python3.10/site-packages/taskiq/brokers/inmemory_broker.py", line 159, in startup
    await maybe_awaitable(handler(self.state))
  File "/app/.venv/lib/python3.10/site-packages/taskiq/utils.py", line 23, in maybe_awaitable
    return await possible_coroutine
  File "/app/deployments/daemon.py", line 44, in app_started
    for stream in SM.active_streams(start_block=settings.APEPAY_START_BLOCK):
  File "/app/.venv/lib/python3.10/site-packages/apepay/__init__.py", line 292, in active_streams
    if stream.is_active:
  File "/app/.venv/lib/python3.10/site-packages/apepay/__init__.py", line 445, in is_active
    return self.time_left.total_seconds() > 0
  File "/app/.venv/lib/python3.10/site-packages/apepay/__init__.py", line 430, in time_left
    return timedelta(seconds=self.contract.time_left(self.creator, self.stream_id))
OverflowError: Python int too large to convert to C int
```

mikeshultz:
This prevents the python sdk from attempting to overflow when instantiating timedeta objects with exceptionally long-funded streams.  The effective limit is like 220k years so even though the accuracy is reduced, humanity will probably have ascended to the stars by that point.

Fixes #65

### DIFF
--- a/sdk/py/apepay/__init__.py
+++ b/sdk/py/apepay/__init__.py
@@ -27,10 +27,9 @@ from .exceptions import (
     TokenNotAccepted,
     ValidatorFailed,
 )
-from .settings import Settings
 from .utils import time_unit_to_timedelta
 
-settings = Settings()
+MAX_DURATION_SECONDS = int(timedelta.max.total_seconds()) - 1
 
 
 class Validator(BaseInterfaceModel):
@@ -431,11 +430,7 @@ class Stream(BaseInterfaceModel):
     @property
     def time_left(self) -> timedelta:
         seconds = self.contract.time_left(self.creator, self.stream_id)
-        return timedelta(
-            seconds=settings.MAX_STREAM_DURATION
-            if seconds > settings.MAX_STREAM_DURATION
-            else seconds
-        )
+        return timedelta(seconds=min(MAX_DURATION_SECONDS, seconds))
 
     @property
     def total_time(self) -> timedelta:
@@ -447,11 +442,7 @@ class Stream(BaseInterfaceModel):
             # NOTE: `last_pull == start_time` if never pulled
             datetime.fromtimestamp(info.last_pull)
             - datetime.fromtimestamp(info.start_time)
-            + timedelta(
-                seconds=settings.MAX_STREAM_DURATION
-                if max_life > settings.MAX_STREAM_DURATION
-                else max_life
-            )
+            + timedelta(seconds=min(MAX_DURATION_SECONDS, max_life))
         )
 
     @property
--- a/sdk/py/apepay/__init__.py
+++ b/sdk/py/apepay/__init__.py
@@ -27,7 +27,10 @@ from .exceptions import (
     TokenNotAccepted,
     ValidatorFailed,
 )
+from .settings import Settings
 from .utils import time_unit_to_timedelta
+
+settings = Settings()
 
 
 class Validator(BaseInterfaceModel):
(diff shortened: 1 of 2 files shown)


sdk/py/apepay/settings.py
```diff
@@ -8,6 +8,7 @@
 class Settings(BaseSettings):
     WARNING_LEVEL: timedelta = timedelta(days=2)
     CRITICAL_LEVEL: timedelta = timedelta(hours=12)
+    MAX_STREAM_DURATION: int = int(timedelta.max.total_seconds()) - 1
```
fubuloubu, line 11: I don't think this needs to be a setting, would just make it a constant in the other file

fubuloubu (approved):

sdk/py/apepay/settings.py
mikeshultz: My thinking was that someone may want to have a max life, but this probably wouldn't be how we'd want to enforce that anyway.  Will change

antazoey (approved):
Was there just some weird test data or something causing this?

sdk/py/apepay/settings.py
fubuloubu: it's sort of a programming language issue

honestly, thinking it should just raise an error for being out of bounds

mikeshultz:
> Was there just some weird test data or something causing this?

I didn't really look into who created it or why (see #65).  It's probably not something we'd see on a prod deployment with USDC because it's only 6 decimals, but I could see it being an accident with an 18 decimal token.

sdk/py/apepay/__init__.py
```diff
@@ -27,10 +27,9 @@
     TokenNotAccepted,
     ValidatorFailed,
 )
-from .settings import Settings
 from .utils import time_unit_to_timedelta
 
-settings = Settings()
```
fubuloubu: I also think settings can likely be removed, shouldn't be a part of the SDK

sdk/py/apepay/settings.py
mikeshultz: Changed.  Let me know if you'd prefer me to throw an error.  If we did that it would still completely break StreamManager, then.

fubuloubu: how do you mean? I think it would make it break on a much clearer error, something like "stream duration out of bounds"

sdk/py/apepay/__init__.py
mikeshultz: I would prefer to only handle the bug in question in this PR.

sdk/py/apepay/settings.py
fubuloubu: I'm picturing this raising a bug on handling that the stream was improperly created, and should not be handled any further

mikeshultz: For instance, iteration of streams via `StreamManager` raises an exception from `time_left()`.  Meaning it's impossible to iterate through streams if one has a bad duration.

mikeshultz: If we just added a different exception, we'd have to handle it (and suppress the stream) in every iterator func by excluding the stream.  Suppression probably isn't the way to go, considering these aren't invalid streams.  This is just a limitation of Python.

PR #68, opened 2023-10-30
fix: onStreamCreate and set default values to token balances [SBK-339]

alexisnsns:
- onstreamcreate function now correctly returns the streams fetched from a streammanager 
- default values are set for token balance and native balance if we cannot fetch them, in order not to interrupt the stream creation

### DIFF
--- a/sdk/js/index.ts
+++ b/sdk/js/index.ts
@@ -284,6 +284,18 @@ export default class StreamManager {
     );
   }
 
+  streamFromEventLog(log: StreamCreated): Stream {
+    return new Stream(
+      this,
+      log.args.creator,
+      log.args.stream_id as number,
+      log.args.token,
+      BigInt(log.args.amount_per_second),
+      this.publicClient,
+      this.walletClient
+    );
+  }
+
   onStreamCreated(
     handleStream: (stream: Stream) => void,
     creator?: Address
@@ -294,25 +306,16 @@ export default class StreamManager {
       eventName: "StreamCreated",
       args: creator ? { creator } : {},
       onLogs: (logs: StreamCreated[]) => {
-        logs
-          .map((log) => {
-            return new Stream(
-              this,
-              log.args.creator,
-              log.args.stream_id as number,
-              log.args.token,
-              BigInt(log.args.amount_per_second),
-              this.publicClient,
-              this.walletClient
-            );
-          })
-          .forEach(handleStream);
+        logs.map((log) => this.streamFromEventLog(log)).forEach(handleStream);
       },
       onError: (error) => console.log(error),
     });
   }
 
-  fetchAllLogs(handleStream: (log: Log) => void, fromBlock?: bigint): void {
+  onAllStreams(
+    handleStream: (stream: Stream) => void,
+    fromBlock?: bigint
+  ): void {
     this.publicClient
       .getContractEvents({
         address: this.address,
@@ -320,8 +323,10 @@ export default class StreamManager {
         eventName: "StreamCreated",
         fromBlock: fromBlock || BigInt(0),
       })
-      .then((logs) => {
-        logs.forEach(handleStream);
+      .then((logs: Log[]) => {
+        (logs as StreamCreated[])
+          .map((log) => this.streamFromEventLog(log))
+          .forEach(handleStream);
       })
       .catch((error) => {
         console.error("Error fetching past logs:", error);
--- a/sdk/js/index.ts
+++ b/sdk/js/index.ts
@@ -235,6 +235,8 @@ export default class StreamManager {
     if (!accountOverride && !this.walletClient?.account)
       throw new Error("Error on create: no account");
 
+    console.log('wallet client', this.walletClient)
+
     const account =
       accountOverride || (this.walletClient?.account?.address ?? "0x0");
     // NOTE: 0x0 shouldn't ever be the value of `account` because of the above error
--- a/sdk/js/index.ts
+++ b/sdk/js/index.ts
@@ -312,7 +312,7 @@ export default class StreamManager {
     });
   }
 
-  fetchAllLogs(callback: (logs: Log[]) => void, fromBlock?: bigint): void {
+  fetchAllLogs(handleStream: (log: Log) => void, fromBlock?: bigint): void {
     this.publicClient
       .getContractEvents({
         address: this.address,
@@ -321,7 +321,7 @@ export default class StreamManager {
         fromBlock: fromBlock || BigInt(0),
       })
       .then((logs) => {
-        callback(logs);
+        logs.forEach(handleStream);
       })
       .catch((error) => {
         console.error("Error fetching past logs:", error);
--- a/sdk/js/index.ts
+++ b/sdk/js/index.ts
@@ -9,7 +9,6 @@ import {
 } from "viem";
 import StreamManagerContractType from "./.build/StreamManager.json";
 
-
 export interface StreamInfo {
   token: Address;
   amount_per_second: bigint;
@@ -260,7 +259,6 @@ export default class StreamManager {
       })) as bigint
     );
 
-
     const hash = await this.walletClient?.writeContract({
       chain: null,
       address: this.address,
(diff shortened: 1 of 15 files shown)


ui/app/src/App.tsx
```diff
@@ -61,8 +67,50 @@ function App() {
...
+            publicClient,
+            walletClient as WalletClient
+          );
+          setSM(newSM);
+        } catch (error) {
+          console.error("Stream Manager Error:", error);
+        }
+      }
+    };
+    fetchStreamManager();
```
fubuloubu, line 99: Use Promise API
```suggestion
    StreamManager.fromAddress(
      config.streamManagerAddress as `0x${string}`,
      publicClient,
      walletClient as WalletClient
    ).then(setSM).catch(console.error);
```

ui/app/src/App.tsx
```diff
@@ -61,8 +67,50 @@ function App() {
...
+
+    // Fetch create streams when SM has been fetched
+    const interval: NodeJS.Timeout = setInterval(() => {
+      if (SM != null && createdStreams.length === 0) {
+        SM.onStreamCreated(handleStreams, address);
```
fubuloubu, line 104: this installs a watcher, so you shouldn't have to do polling because it should do it already (hopefully via websockets)

fubuloubu: in which case, we can probably chain off the promise above and make the `.then` handler a bit more complex to install this when it sets the `SM` state variable

ui/app/src/App.tsx
```diff
@@ -61,8 +67,50 @@ function App() {
...
   };
 
-  // Get stream from CreateStream to pass it as props
+  const publicClient = usePublicClient();
+  const walletClient = useWalletClient()?.data;
+  const { address } = useAccount();
+
   const [stream, setStream] = useState<Stream | null>(null);
+  const [SM, setSM] = useState<StreamManager | null>(null);
+  const [createdStreams, setCreatedStreams] = useState<Stream[]>([]);
```
fubuloubu, line 76: would have it manage all streams from the contract, and then add a method to the JS SDK to fetch all past streams to start with

when you fetch the `StreamManager` I would run through all past events to catch up to present, and then run the `.onStreamCreated` watcher to listen to new streams created

sdk/js/index.ts
```diff
@@ -2,7 +2,6 @@ import {
   Abi,
   Address,
   ByteArray,
-  Log,
```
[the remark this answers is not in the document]
alexisnsns: this topics thing is confusing; I'm not the first one to struggle here https://github.com/wagmi-dev/viem/discussions/981 ; I'll have another look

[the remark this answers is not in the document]
alexisnsns: Ok so the way I found to make it work was to extend the viem Log type to use args; otherwise you can't use log.args and this is needed to get the stream params (token, streamid etc)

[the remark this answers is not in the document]
fubuloubu: Oh neat!

sdk/js/index.ts
```diff
@@ -272,32 +283,34 @@ export default class StreamManager {
...
+    try {
+      this.publicClient.watchContractEvent({
+        address: this.address,
+        abi: StreamManagerContractType.abi as Abi,
+        eventName: "StreamCreated",
+        args: creator ? { creator } : {},
+        onLogs: (logs: Log[]) => {
+          (logs as StreamCreated[])
+            .map((log) => {
+              const stream = log.args;
```
fubuloubu, line 298: So this line no longer needed

sdk/js/index.ts
```diff
@@ -272,32 +283,34 @@ export default class StreamManager {
...
-      onLogs,
-    });
+    try {
+      this.publicClient.watchContractEvent({
+        address: this.address,
+        abi: StreamManagerContractType.abi as Abi,
+        eventName: "StreamCreated",
+        args: creator ? { creator } : {},
+        onLogs: (logs: Log[]) => {
+          (logs as StreamCreated[])
```
fubuloubu, line 296: Wonder if this would work
```suggestion
        onLogs: (logs: StreamCreated[]) => {
          logs
```

alexisnsns: it does!

sdk/js/index.ts
```diff
@@ -52,9 +63,9 @@ export class Stream {
     publicClient: PublicClient,
     walletClient?: WalletClient
   ): Promise<Stream> {
-    const creator = log.topics[2] as Address;
-    const streamId = Number(log.topics[3]);
-    const token = log.topics[4] as Address;
+    const creator = ("0x" + (log.topics[2] as string).slice(-40)) as Address;
+    const streamId = isNaN(Number(log.topics[3])) ? 1 : Number(log.topics[3]);
```
fubuloubu, line 67: when it's NaN, I think it should raise

sdk/js/index.ts
```diff
@@ -272,32 +283,43 @@ export default class StreamManager {
...
+      onError: (error) => console.log(error),
     });
   }
+
+  async fetchAllLogs(callback: (logs: Log[]) => void) {
+    try {
+      const logs = await this.publicClient.getLogs({
+        address: this.address,
+        // TODO: pass block as function params? Otherwise takes too long
+        fromBlock: 4596186n,
```
fubuloubu, line 318: yeah, we have this in the python sdk, can have this be an optional parameter that defaults to block 0, and then can see if you can obtain the deployment block of the StreamManager somehow

sdk/js/index.ts
```diff
@@ -272,32 +283,43 @@ export default class StreamManager {
...
+              this.walletClient
+            );
+          })
+          .forEach(handleStream);
+      },
+      onError: (error) => console.log(error),
     });
   }
+
+  async fetchAllLogs(callback: (logs: Log[]) => void) {
```
fubuloubu, line 313: maybe use the same `handleStream` type function handle, since the getLogs could probably be using batches

sdk/js/index.ts
alexisnsns: With the change from fetchLogs to getcontractevents (using eventName: "StreamCreated") we dont get streamID issues, as the logs are already cleaned

sdk/js/index.ts
alexisnsns: Using getContractevents now

sdk/js/index.ts
fubuloubu: probably need to find some way to fetch this during the async constructor method

sdk/js/index.ts
```diff
@@ -52,9 +63,9 @@ export class Stream {
     publicClient: PublicClient,
     walletClient?: WalletClient
   ): Promise<Stream> {
-    const creator = log.topics[2] as Address;
+    const creator = ("0x" + (log.topics[2] as string).slice(-40)) as Address;
```
fubuloubu, line 66: Is this necessary since it's parsed as an address?

sdk/js/index.ts
```diff
@@ -272,32 +285,44 @@ export default class StreamManager {
...
+              this.walletClient
+            );
+          })
+          .forEach(handleStream);
+      },
+      onError: (error) => console.log(error),
     });
   }
+
+  async fetchAllLogs(callback: (logs: Log[]) => void, fromBlock?: bigint) {
```
fubuloubu, line 315: We can probably switch this into the promise API using the handler function so it doesn't need to be async (since it doesn't return anything)

sdk/js/index.ts
alexisnsns: Yes otherwise you get a couple of duplicate '0' after the 'Ox'

fubuloubu: oh, I mean like why does `const creator = log.topics[2] as Address` not work?

alexisnsns: because log topics returns this before you clean it, so it cannot be used as it is. Address type is just `0x${string}` so it doesnt count the number of characters
Screenshot 2023-11-02 at 12 07 15

sdk/js/index.ts
```diff
@@ -312,17 +312,19 @@ export default class StreamManager {
     });
   }
 
-  async fetchAllLogs(callback: (logs: Log[]) => void, fromBlock?: bigint) {
-    try {
-      const logs = await this.publicClient.getContractEvents({
+  fetchAllLogs(callback: (logs: Log[]) => void, fromBlock?: bigint): void {
```
fubuloubu, line 315: was hoping to refactor like this
```suggestion
  fetchAllLogs(handleStream: (log: Log) => void, fromBlock?: bigint): void {
```

sdk/js/index.ts
```diff
@@ -272,32 +285,46 @@ export default class StreamManager {
...
+    this.publicClient
+      .getContractEvents({
+        address: this.address,
+        abi: StreamManagerContractType.abi as Abi,
+        eventName: "StreamCreated",
+        fromBlock: fromBlock || BigInt(0),
+      })
+      .then((logs) => {
+        callback(logs);
```
fubuloubu, line 324:
```suggestion
        logs.forEach(handleStream);
```

fubuloubu (approved):

PR #71, opened 2023-11-06
fix: context loss

alexisnsns:
Context seems to be lost when calling streamFromEventLog (error: this is undefined) in the SDK. 
Patch to bind the context when calling the function.

### DIFF
--- a/sdk/js/index.ts
+++ b/sdk/js/index.ts
@@ -284,7 +284,12 @@ export default class StreamManager {
     );
   }
 
-  streamFromEventLog(log: StreamCreated): Stream {
+  streamFromEventLog = (log: StreamCreated): Stream => {
+
+    // TODO: remove console.logs
+    console.log('this in streamfromeventlog', this)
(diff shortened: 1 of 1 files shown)


alexisnsns:
console.logs in index.ts are used to debug when using the SDK. They need to be removed in a later release.

sdk/js/index.ts
```diff
@@ -317,6 +321,10 @@ export default class StreamManager {
     fromBlock?: bigint,
     toBlock?: bigint
   ): void {
+    // TODO: remove console.logs
```
alexisnsns, line 324: To be removed in a later release

sdk/js/index.ts
```diff
@@ -284,7 +284,11 @@ export default class StreamManager {
     );
   }
 
-  streamFromEventLog(log: StreamCreated): Stream {
+  streamFromEventLog = (log: StreamCreated): Stream => {
+    // TODO: remove console.logs
```
alexisnsns, line 288: To be removed in a later release

fubuloubu (approved):

PR #73, opened 2023-11-07
Fix/localinstall [SBK-348]

alexisnsns:
Import apeworx/apepay and apeworx/apepay-react in the demo app by doing a local install instead of doing local imports

### DIFF
--- a/ui/app/src/App.tsx
+++ b/ui/app/src/App.tsx
@@ -5,10 +5,6 @@ import config from "./config";
 // NOTE: Do this or else it won't render (or create your own CSS)
 import "rc-slider/assets/index.css";
 import "./styles.css";
-// import CreateStream from "lib/CreateStream";
-// import StreamStatus from "lib/StreamStatus";
-// import CancelStream from "lib/CancelStream";
-// import UpdateStream from "lib/UpdateStream";
 import StreamManager, { Stream } from "@apeworx/apepay";
 import  { CreateStream, CancelStream, UpdateStream, StreamStatus } from "@apeworx/apepay-react";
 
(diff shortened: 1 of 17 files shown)


fubuloubu (approved):

ui/app/src/App.tsx
```diff
@@ -94,7 +98,7 @@ function App() {
 
   // Fetch logs starting from this block
   // TODO: find a way to get the SM deployment block
-  const fromBlock = 4615000n;
+  const fromBlock = config.fromBlock ? BigInt(config.fromBlock) : undefined;
```
fubuloubu, line 101: Wouldn't null be better?

fubuloubu (approved):

ui/app/src/App.tsx
alexisnsns: The expected type in getContractEvents is `fromBlock?: bigint | BlockTag | undefined`, TS doesn't like it when i replace it with null

fubuloubu: oh, is that what `?` means? a type or `undefined`?

alexisnsns: yes that's why the '?' in the optional param; if you still pass an optional param when calling the function (as in our case), you'll need to pass it as undefined

ui/lib/UpdateStream.tsx
```diff
@@ -160,7 +160,7 @@ const UpdateStream: React.FC<UpdateStreamProps> = (props) => {
             disabled={isButtonDisabled}
             className="update-stream-button"
           >
-            Fund stream
+            Fund
```
fubuloubu, line 163: I think `Add Time` might be nicer

fubuloubu (approved):

fubuloubu (approved):

PR #75, opened 2023-11-16
feat: implement routing in the demo app [SBK-362]

alexisnsns:
Following @fubuloubu suggestion:

    /{StreamManager-address} - display all streams for a particular stream manager address (and load new ones as created), add filter params to filter down streams

    /{StreamManager-address}/{creator-address} - display all streams for a particular stream manager, by a particular creator (same as above, load new ones as they come in and add filter params)

    /{StreamManager-address}/create - Create a new Stream, re-route to detail view once created (using registerStream prop)

    /{StreamManager-address}/{creator-address}/{stream-id} - display the details of a particular stream

Implemented these routes so that there is now a separation of concerns, not everything is done on the main page; an outsider can understand the streamManager/creator/stream logic more easily now; React components are used in only one file now so that it is easier to make sense of their logic.


ui/app/src/App.tsx
```diff
@@ -246,6 +246,18 @@ function App() {
                 <p> Creator: {selectedStream.creator}</p>
                 <p> Amount per second: {String(streamInfo.amountPerSecond)}</p>
                 <p> Funded amount: {String(streamInfo.fundedAmount)}</p>
+                <p>
+                  <Link
+                    to={`/${selectedStream.creator}/${selectedStream.streamId}`}
```
fubuloubu, line 251: Does this need stream manager address too?

ui/app/src/CreatePage.tsx
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+import { useParams } from "react-router-dom";
+import BackButton from "./BackButton";
+
+const StreamPage = () => {
+  const { sm } = useParams();
+
+  return (
+    <>
+      <h1>Create Stream on {sm}</h1>
```
fubuloubu, line 10: Can migrate `CreateStream` component here

ui/app/src/App.tsx
alexisnsns: lol this PR is not ready at all that's why I left it as draft

alexisnsns: still plenty to be done before it's ready for review!

alexisnsns:
@fubuloubu as this PR is mostly about routing and UI, here is a quick overview if you don't want to test it locally:

on /{StreamManager-address}, you can see all the streams of a specific streamManager, filtered by creator; you can then click on a creator to see its streams, or directly on a stream if you know what you're looking for, or click on a button to create a new stream:
Screenshot 2023-11-17 at 18 34 59

Then if you click on a creator you'll see all his streams on a given streamManager ( /{StreamManager-address}/{creator-address}:
Screenshot 2023-11-17 at 18 39 00

Then you can display the details of a particular stream (/{StreamManager-address}/{creator-address}/{stream-id})
Screenshot 2023-11-17 at 18 34 47

otherwise from the streamManager page you have a button to create a stream (/{StreamManager-address}/create):

Screenshot 2023-11-17 at 18 44 21

fubuloubu (approved):
This is a great update! I think it is good to merge as is. Eventually we want to host it and we can develop it up more over time.

One suggestion is that if we could do a table-like list of streams, that could be sorted or filtered it would probably make navigation easier

PR #78, opened 2023-11-22
refactor: make isCancellable sync [SBK-370]

alexisnsns:
isCancellable used to be async, with the cancel component displaying a loading message while waiting for the promise to resolve, which would take ~5-10 secs.

 I added the startTime in the Stream constructor to make it sync.

Now there is little to no loading time in the cancel stream component, which is a great UX win. 

Screenshot 2023-11-22 at 12 03 57

It allows to display the remaining time almost instantly when you load the stream.


alexisnsns:
Also changed wording to add 'simulations' and edited the button css to make it so that on the demo app the user knows a button is disabled

alexisnsns:
Screenshot 2023-11-22 at 13 57 52

Also added a favicon as every error I can make disappear is a win and helps js debugging more easily as the console becomes more readable

ui/lib/CreateStream.tsx
```diff
@@ -135,7 +135,7 @@ const CreateStream = (props: CreateStreamProps) => {
     (tokenData?.value || BigInt(0)) / BigInt(props.amountPerSecond),
   );
 
-  // TODO: Increase stability of deployments beyond a week
+  // TODO: Increase stability of simulations beyond a week
   const maxTimeDays: number = Math.min(Math.floor(maxTime / SECS_PER_DAY), 7); // Up to a week
```
fubuloubu, line 139: hmm, why is there a limit here? the limit is driven by chaosnet use case

ui/lib/CancelStream.tsx
```diff
@@ -62,42 +57,12 @@ const CancelStream: React.FC<CancelStreamProps> = (props) => {
...
-        </div>
-      ) : !isButtonEnabled &&
-        !inProgress &&
-        startTime !== 0 &&
-        Number(timeBeforeCancellability) !== 0 ? (
+      {!isButtonEnabled && !inProgress && timeBeforeCancellability > 0 ? (
         <>
           <div className="cancel-stream-label-min-life">
-            Deployment cannot be cancelled yet: its minimum life is
+            Simulation cannot be cancelled yet: its minimum life is
```
fubuloubu, line 65: I think this label should be configurable (maybe default is this)
```suggestion
            Stream cannot be cancelled yet: its minimum life is
```

ui/app/src/StreamPage.tsx
```diff
@@ -204,7 +212,7 @@ const StreamPage = () => {
               {/* UpdateStream callback */}
               {updateStatus && (
                 <p className="label-close-modal">
-                  -Deployment is being updated- Close modal
+                  -Simulation is being updated- Close modal
```
fubuloubu, line 215: same here

sdk/js/index.ts
fubuloubu, line 1: This is a smart update

fubuloubu (approved):
Need some updates for language here

ui/lib/CreateStream.tsx
alexisnsns: The comment was already there, I just changed the wording. I'll remove the comment then 👍

PR #99, opened 2024-08-27
fix: use log-level fix from ape

antazoey:
seeing if my loglevel fix works, and then if it does, can replace with ape 0.8.14

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -35,8 +35,7 @@ jobs:
       - uses: ApeWorX/github-action@v3
       - run: ape compile --force --size
       # Upload ape build files for testing
-      - uses: actions/upload-artifact@v3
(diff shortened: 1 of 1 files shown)


fubuloubu:
Verified, https://github.com/ApeWorX/ape/pull/2249 merged but not released

.github/workflows/test.yaml
```diff
@@ -50,15 +50,16 @@ jobs:
       - uses: actions/checkout@v4
       - uses: foundry-rs/foundry-toolchain@v1
       - uses: ApeWorX/github-action@v3
+        with:
+          ape-version-pin: git+https://github.com/antazoey/ape.git@fix/debug-str-err
```
fubuloubu, line 54:
```suggestion
```

fubuloubu (approved):

PR #100, opened 2024-09-05
fix(contract): bug with ordering of operations

fubuloubu:
There was a bug in the StreamManager implementation concerning some order of operations that would allow theft to occur w/ callback tokens

### DIFF
--- a/contracts/StreamManager.vy
+++ b/contracts/StreamManager.vy
@@ -223,12 +223,11 @@ def cancel_stream(
     assert self.streams[creator][stream_id].start_time + MIN_STREAM_LIFE <= block.timestamp
 
     funded_amount: uint256 = self.streams[creator][stream_id].funded_amount
-    amount_locked: uint256 = funded_amount  - self._amount_unlocked(creator, stream_id)
+    amount_locked: uint256 = funded_amount - self._amount_unlocked(creator, stream_id)
+    self.streams[creator][stream_id].funded_amount = funded_amount - amount_locked
 
     token: ERC20 = self.streams[creator][stream_id].token
     assert token.transfer(creator, amount_locked, default_return_value=True)
-
-    self.streams[creator][stream_id].funded_amount = funded_amount - amount_locked
 
     log StreamCancelled(creator, stream_id, amount_locked, reason)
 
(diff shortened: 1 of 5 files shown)


contracts/StreamManager.vy
```diff
@@ -223,13 +223,12 @@ def cancel_stream(
     assert self.streams[creator][stream_id].start_time + MIN_STREAM_LIFE <= block.timestamp
 
     funded_amount: uint256 = self.streams[creator][stream_id].funded_amount
-    amount_locked: uint256 = funded_amount  - self._amount_unlocked(creator, stream_id)
+    amount_locked: uint256 = funded_amount - self._amount_unlocked(creator, stream_id)
+    self.streams[creator][stream_id].funded_amount = funded_amount - amount_locked
```
antazoey, line 227: Is there a case where `funded_amount` is 0 and `amount_locked` is also 0, so like canceling an empty stream, maybe by accident ( same operation twice in a block?).
Should be assert the stream is funded before canceling, I guess is what im wondering

fubuloubu: [ce90246](https://github.com/ApeWorX/ApePay/pull/100/commits/ce90246a521c814ea8807092bed7818de03268e8)

contracts/StreamManager.vy
```diff
@@ -238,17 +242,13 @@ def cancel_stream(
 @external
 def claim(creator: address, stream_id: uint256) -> uint256:
     funded_amount: uint256 = self.streams[creator][stream_id].funded_amount
-    claimed_amount: uint256 = min(
-        self._amount_unlocked(creator, stream_id),
-        funded_amount,
-    )
+    claim_amount: uint256 = self._amount_unlocked(creator, stream_id)
```
antazoey, line 245: Similarly should we revert on a zero amount to avoid tx costs?

tests/conftest.py
```diff
@@ -27,6 +24,20 @@ def tokens(create_token, payer, request):
     return [create_token(payer) for _ in range(int(request.param.split(" ")[0]) + 1)]
 
 
+@pytest.fixture(scope="session")
+def token(tokens):
+    if len(tokens) == 0:
+        pytest.skip("No valid tokens")
+
+    return tokens[0]
```
antazoey, line 32: Perf on the test would be to use try/ except because i think token lists len can be slow sometimes

On except, do the pytest.skip

antazoey (approved):

PR #108, opened 2024-09-19
refactor: drop poetry and use newer package manifest loader

fubuloubu:
### What I did

Was a little annoyed using poetry, this refactors back to using it with standard format tools (also works w/ `uv pip`)

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -33,6 +33,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: ApeWorX/github-action@v3
+        with:
+          python-version: "3.11"
       - run: ape compile --force --size
       # Upload ape build files for testing
       - uses: actions/upload-artifact@v4
@@ -52,6 +54,8 @@ jobs:
       - uses: actions/checkout@v4
       - uses: foundry-rs/foundry-toolchain@v1
       - uses: ApeWorX/github-action@v3
+        with:
+          python-version: "3.11"
       # Ape-built artifacts used for testing
       - uses: actions/download-artifact@v4
         with:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -57,7 +57,6 @@ jobs:
         with:
           name: ape-build-folder
           path: .build/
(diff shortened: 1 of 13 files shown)


fubuloubu:
forgot to add the work I had refactoring to better ways of using package resources loader

.github/workflows/test.yaml
```diff
@@ -52,12 +54,13 @@ jobs:
       - uses: actions/checkout@v4
       - uses: foundry-rs/foundry-toolchain@v1
       - uses: ApeWorX/github-action@v3
```
mikeshultz, line 56: question: wait, what is this?

pyproject.toml
```diff
@@ -1,34 +1,54 @@
...
-black = "^24.4.2"
+[tool.setuptools.package-data]
+apepay = [ "manifest.json", "py.typed" ]
 
-[build-system]
-requires = ["poetry-core"]
-build-backend = "poetry.core.masonry.api"
+[tool.setuptools_scm]
+# NOTE: Config entry needed for this plugin to function
+version_file = "sdk/py/apepay/version.py"
```
mikeshultz, line 47: suggestion: You should be able to get things running without this file. Unless this file is used within the SDK or something.

If it complains about not being able to find version, you might need to set the `root`. For example:

```toml
[tool.setuptools_scm]
root = "../.."
```

Assuming it thinks `sdk/py` is the root.

mikeshultz (approved):
Bit of commentary. Nothing blocking.  LGTM

.github/workflows/test.yaml
fubuloubu: we have a github action!
https://github.com/ApeWorX/github-action

pyproject.toml
fubuloubu: ah okay, yeah this seems like a better way to configure. will do

fubuloubu: okay, yeah this isn't even needed

I was having issues with using `tool.setuptools-scm` as a key, which never works, and that was one attempt I made at resolving that

removing

.github/workflows/test.yaml
mikeshultz: This is cool

PR #109, opened 2024-09-19
refactor(scripts): use multicall w/ claim

fubuloubu:
### What I did

Small refactor of the manage script to take advantage of multicall on networks where it is available

### Checklist

- [x] All changes are completed
~~- [ ] New test cases have been added~~
- [x] Documentation has been updated

### DIFF
--- a/scripts/manage.py
+++ b/scripts/manage.py
@@ -1,7 +1,7 @@
 import click
+from ape.cli import ConnectedProviderCommand, account_option, network_option
 from ape.types import AddressType
 from ape_ethereum import multicall
-from ape.cli import ConnectedProviderCommand, network_option, account_option
 from apepay import StreamManager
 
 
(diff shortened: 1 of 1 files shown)


scripts/manage.py
```diff
@@ -33,12 +33,25 @@ def unclaimed(network, start_block, address):
...
+    if not use_multicall:
+        for _ in range(batch_size):
+            try:
+                stream = next(unclaimed_streams)
+            except StopIteration:
+                return
+
+            stream.claim(sender=account)
+
+        raise click.UsageError("More claims needed")
```
mikeshultz, line 53: suggestion(non-blocking): Maybe change this to something like "There are more unprocessed claims"? Or maybe even not raise an exception at all, just exit clean.

mikeshultz (approved):

scripts/manage.py
fubuloubu: Yeah, I think no need to raise, but should just give a a nice reminder that there are more claims to do

fubuloubu: 22c9527

Issue #107, opened 2024-09-19
Pricing Validator example


fubuloubu:
### Describe

Show how you would put a pricing model validator into a StreamManager

### Dependencies

n/a

Issue #103, opened 2024-09-13
Allowlist and Denylist example validators


fubuloubu:
`Allowlist.vy`:
```vyper
owner: public(address)
is_allowed: public(HashMap[address, bool])

event Allowed:
    user: indexed(address)

event Denied:
    user: indexed(address)

@deploy
def __init__(allowed: DynArray[address, 100]):
    self.owner = msg.sender
    for user in allowed:
        self.is_allowed[user] = True
        log Allowed(user)

@external
def allow(allowed: DynArray[address, 100]):
    assert msg.sender == self.owner

    for user in allowed:
        self.is_allowed[user] = True
        log Allowed(user)

@external
def deny(denied: DynArray[address, 100]):
    assert msg.sender == self.owner

    for user in denied:
        self.is_allowed[user] = False
        log Denied(user)

@external
def validate(creator: address, ...):
    assert self.is_allowed[creator]
    return MAX_UINT256
```

`Denylist.vy`:
```vyper
from .. import StreamManager

owner: public(address)
is_denied: public(HashMap[address, bool])

event Allowed:
    user: indexed(address)

event Denied:
    user: indexed(address)

@deploy
def __init__(denied: DynArray[address, 100]):
    self.owner = msg.sender
    for user in allowed:
        self.is_denied[user] = True
        log Allowed(user)

@external
def allow(allowed: DynArray[address, 100]):
    assert msg.sender == self.owner

    for user in allowed:
        self.is_denied[user] = False
        log Allowed(user)

@external
def deny(denied: DynArray[address, 100]):
    assert msg.sender == self.owner

    for user in denied:
        self.is_denied[user] = True
        log Denied(user)

@external
def cancel_stream(manager: StreamManager, creator: address, stream_id: uint256):
    # NOTE: Batch-able via Multicall3
    assert self.is_denied[creator]
    manager.cancel_stream(creator, stream_id)

@external
def validate(creator: address, ...):
    assert not self.is_denied[creator]
    return MAX_UINT256
```

Issue #72, opened 2023-11-07
Refactor stream fetching in the SDK [SBK-347]


alexisnsns:
When calling onAllStreams, fetching the streams (using streammanager.publicClient.getContractEvents) can be slow on public nets if you don't pass a 'fromBlock' param to start fetching only from a certain block.

The way we fetch the past streams can be improved.

- There's a linear increment in how the stream works, so you could just iterate them in batches using multicall for speed
- We can add a native multicall-like search for live streams into the contract (maybe query 256 at a time)
- Use the [Graph](https://thegraph.com/)?

PR #112, opened 2024-09-25
refactor(contracts)!: StreamManager upgrades

fubuloubu:
### What I did

- `Stream.stream_id -> Stream.id`
- reordering/renaming of some event arguments
- reordering of `create_stream` function
- remove unused functions
- adds Allowlist/Denylist validators
- adds PricingExample validator
- allow StreamManager.controller to transition ownership
- allow Stream.owner to transition ownership

fixes: #103
fixes: #104
fixes: #105
fixes: #107 
partial fix: #72 

### How I did it

some larger refactors to contracts, examples, py sdk, etc.

NOTE: No update to js-sdk

### Checklist

- [x] All changes are completed
- [x] New test cases have been added
- [x] Documentation has been updated


fubuloubu:
[ca8c85b](https://github.com/ApeWorX/ApePay/pull/112/commits/ca8c85bd72cb6c84c60b6ae22557f07cceaa9fd5) related: https://github.com/ApeWorX/ape/pull/2277

sdk/py/apepay/manager.py
```diff
@@ -89,21 +91,32 @@ def _parse_validator(self) -> Callable[[_ValidatorItem], Validator]:
...
-        return cast(ContractTransactionHandler, order_validators)
+        return cast(ContractTransactionHandler, set_validators)
 
     def add_validators(self, *new_validators: _ValidatorItem, **txn_kwargs) -> ReceiptAPI:
         return self.set_validators(
             *(set(self.validators) | set(map(self._parse_validator, new_validators))),
             **txn_kwargs,
         )
 
+    def replace_validator(self, old_validator, new_validator, **txn_kwargs) -> ReceiptAPI:
```
antazoey, line 111: Missing param types?

sdk/py/apepay/manager.py
```diff
@@ -89,21 +91,32 @@ def _parse_validator(self) -> Callable[[_ValidatorItem], Validator]:
...
+        return self.set_validators(
```
antazoey, line 112: How come this method doesn't take already parsed `_ValidatorItem`s like other alike methods, such as `set_validator()`

sdk/py/apepay/manager.py
```diff
@@ -89,21 +91,32 @@ def _parse_validator(self) -> Callable[[_ValidatorItem], Validator]:
...
         @wraps(self.contract.set_validators)
-        def order_validators(*validators: _ValidatorItem, **txn_kwargs) -> ReceiptAPI:
+        def set_validators(*validators: _ValidatorItem, **txn_kwargs) -> ReceiptAPI:
             # NOTE: Always keep sets sorted, ensure no duplicates
-            return self.contract.set_validators(
-                sorted(v.address for v in set(map(self._parse_validator, validators))),
-                **txn_kwargs,
-            )
+            new_validators = sorted(v.address for v in set(map(self._parse_validator, validators)))
+            # NOTE: Okay to use Click here, intended primarily to support interactive features
```
antazoey, line 97: Does this comment mean you shouldn't use this method in a regular library sort of way?

sdk/py/apepay/manager.py
antazoey, line 111: a test for the `replace_validator()` method might be nice

sdk/py/apepay/manager.py
```diff
@@ -89,21 +91,32 @@ def _parse_validator(self) -> Callable[[_ValidatorItem], Validator]:
...
+            click.echo("Validators update:")
+            for line in Differ().compare(tuple(v.address for v in self.validators), new_validators):
```
antazoey, line 99: TIL I learned about `Differ`!

antazoey (approved):
approved w/ thoughts

contracts/StreamManager.vy
```diff
@@ -17,243 +16,313 @@
...
+    self.validators = validators
+
 
 @external
-def set_validators(validators: DynArray[Validator, MAX_VALIDATORS]):
-    assert msg.sender == self.owner
-    self.validators = validators
+def transfer_control(new_controller: address):
+    # NOTE: can revoke transfer at any time calling this method with `self.controller`
+    assert msg.sender == self.controller
```
antazoey, line 128: Should we add a dev-message here? (and other similar spots)

contracts/Validator.vyi
```diff
@@ -1,13 +1,43 @@
...
+    @param token The ERC20 token that is being used to fund the stream
+    @param amount The amount of `token` that is funding the stream. Can be used to ensure product
+        time limits are within acceptable bounds by computing a rate from the subtotal price for
+        the subset of matching products this validator handles.
+    @param products The array of product codes or reasons for the triggering of the action. Note
+        that the encoding of this value is entirely application-specific, and multiple supported
+        encodings could be potentially be supported by other connected Validators.
+    @return uint256 streaming rate for the subset of products this validator handles. Used to
+        compute the total streaming rate for all products by summing all subtotals of all
+        connected validators to the Streammanager.
```
antazoey, line 41:
```suggestion
        connected validators to the StreamManager.
```

contracts/validators/Allowlist.vy
```diff
@@ -0,0 +1,52 @@
...
+event Denied:
+    user: indexed(address)
+
+
+@deploy
+def __init__(allowed: DynArray[address, 100]):
+    self.owner = msg.sender
+    for user: address in allowed:
+        self.is_allowed[user] = True
+        log Allowed(user)
```
antazoey, line 23: would it be cheaper to log once with an array of all users?

sdk/py/apepay/manager.py
```diff
@@ -87,123 +91,123 @@ def _parse_validator(self) -> Callable[[_ValidatorItem], Validator]:
...
+        amount: int,
+        products: list[HexBytes],
+    ) -> int:
+        total = 0
+
+        for validator in self.validators:
+            try:
+                total += validator(funder, token, amount, products)
+            except ContractLogicError:
+                pass
```
antazoey, line 153:
```suggestion
                continue
```

would do better with future potential changes (but otherwise no different in logic)

sdk/py/apepay/manager.py
```diff
@@ -87,123 +91,123 @@ def _parse_validator(self) -> Callable[[_ValidatorItem], Validator]:
...
-        if reason is not None:
-            if isinstance(reason, dict):
-                reason = json.dumps(reason, separators=(",", ":"))
+        if not isinstance(amount, int):
+            amount = self.conversion_manager.convert(amount, int)
+        assert isinstance(amount, int)  # for mypy
 
-            if isinstance(reason, str):
-                reason = reason.encode("utf-8")
+        args: list[Any] = [token, amount, products]
```
antazoey, line 172: Having a local variable named `args` is something I try to avoid because `pdb` also has something called `args` so when debugging using `pdb`, it is harder to inspect the variable

sdk/py/apepay/streams.py
```diff
@@ -20,139 +17,94 @@
...
+        seconds = self.contract.time_left(self.id)
+        assert seconds < MAX_DURATION_SECONDS, "Invaraint wrong"
+        return timedelta(seconds=seconds)
 
     @property
     def total_time(self) -> timedelta:
-        info = self.info  # NOTE: Avoid calling contract twice
-        # NOTE: Measure time-duration of unclaimed amount remaining (locked and unlocked)
-        max_life = int(info.funded_amount / info.amount_per_second)
+        info = self.info  # NOTE: Avoid calling contract twice by caching
```
antazoey, line 97: this comment is exactly why I used to gripe about putting calls in properties

sdk/py/apepay/streams.py
```diff
@@ -20,139 +17,94 @@
...
+
+        # NOTE: Measure time-duration of unclaimed amount remaining
+        remaining_life = int(info.funded_amount / info.amount_per_second)
```
antazoey, line 100: i think you could just do this:

```suggestion
        remaining_life = info.funded_amount // info.amount_per_second
```

(two slashes)

sdk/py/apepay/streams.py
```diff
@@ -20,139 +17,94 @@
...
+        assert remaining_life < MAX_DURATION_SECONDS, "Invariant wrong"
```
antazoey, line 101: may want an exception here

antazoey (commented):
oops more comments

antazoey:
something weird I noticed, can't tell if it because of Ape or the ApePay library, but if you run the tests w/o contracts being compiled, it will compile them, but then all tests will fail with `'Project' object has no attribute 'StreamManager'`

sdk/py/apepay/streams.py
fubuloubu: This is a usage invariant that should never be violated, but I wanted it here for when you do testing to make sure it raises properly (`remaining_life` is time in seconds, violating this invariant means you want a stream to run for the rest of the life of the universe)

Basically only if your streaming rates are very small and your funding rates are very large would it be a problem

antazoey: Ah ok. Got it. However, my experience working on the framework -- if there is a hacky bad thing people can do, they will somehow do it.

fubuloubu: It will raise a weirder error about C types being out of bounds

mikeshultz (commented):
I did a quick pass. Want to come back and do another a little later.

suggestion: document more. Maybe file-level docstrings in the header, or maybe in the README.

question: What exactly is a product? Are there any constraints?  If it's just up to validators what exactly that means to the contract, maybe that should be explained a bit more.

Looking at the demo it leaves me a bit confused:

```python
>>> HashBytes32(b"\x00" * 24 + b"\x01" + b"\x00" * 7)
HexBytes('0x0000000000000000000000000000000000000000000000000100000000000000')
>>> 0x0000000000000000000000000000000000000000000000000100000000000000
72057594037927936
>>> 0x0000000000000000000000000000000000000000000000000100000000000000 // 1e18
0.0
```

Even if you adjust for different decimals for a token, I'm not sure how this matches the comment of "256 tokens/hour".  Feel like I'm missing something.

fubuloubu:
> question: What exactly is a product? Are there any constraints? If it's just up to validators what exactly that means to the contract, maybe that should be explained a bit more.

Yes, it is up to the validator. :+1: on better docs/explanation there (but right now it's more for us)

> Even if you adjust for different decimals for a token, I'm not sure how this matches the comment of "256 tokens/hour". Feel like I'm missing something.

The "TestValidator" in there basically just converts the product code directly into the streaming rate, so that hex number is `0.072057594 tokens/second` or 259.4 tokens/hour ( a little bit off)

> ```python
> >>> 0x0000000000000000000000000000000000000000000000000100000000000000 // 1e18
> 0.0
> ```

Don't forget that the token decimals are really only for display purposes only

```python
>>> 0x0000000000000000000000000000000000000000000000000100000000000000 / 1e18
0.07205759403792794
```

sdk/py/apepay/manager.py
fubuloubu: Just added it, felt useful. Will add

fubuloubu: 9bbd0e5

sdk/py/apepay/manager.py
fubuloubu: 9bbd0e5

sdk/py/apepay/manager.py
fubuloubu: 9bbd0e5

contracts/StreamManager.vy
```diff
@@ -17,243 +16,313 @@
...
-            extcall validator.validate(msg.sender, token, amount_per_second, reason),
-        )
 
-    assert max_stream_life >= funded_amount // amount_per_second  # dev: max stream life too small
+@external
+def set_token_accepted(token: IERC20, is_accepted: bool):
+    if Ability.MODFIY_TOKENS not in self.capabilities[msg.sender]:
+        assert msg.sender == self.controller  # dev: insufficient capability
+
+    self.token_is_accepted[token] = is_accepted
```
mikeshultz, line 165: thought: There's a potential foot gun here by accepting tokens with different decimals. Presumably that could be possible with an intelligent validator but this could trip up users.

sdk/py/apepay/manager.py
fubuloubu: You can, it would just probably not be common and you would get some extra printout happening

Really, I just couldn't figure out the best way to use Ape's logger via `ManagerAccessMixin`

contracts/StreamManager.vy
fubuloubu: really just one spot, all the rest rely on the capability fallback onto "is this the controller?" and hence this is different

fubuloubu: 080ef47

contracts/validators/Allowlist.vy
fubuloubu: oh yeah probably

sdk/py/apepay/manager.py
fubuloubu: 8627d93

contracts/validators/Allowlist.vy
fubuloubu: well... actually it's not as nice, if you are indexing on a particular user, since indexed arrays actually get hashed (so you can't index on a particular element in an array existing)

contracts/StreamFactory.vy
```diff
@@ -16,24 +16,35 @@ BLUEPRINT: public(immutable(address))
...
+    min_stream_time: uint256 = ONE_HOUR,
+) -> address:
+    deployment: address = create_from_blueprint(  # dev: only one deployment allowed
         BLUEPRINT,
         msg.sender,  # Only caller can create
-        ONE_HOUR,  # Safety parameter (not configurable)
+        min_stream_time,  # Safety parameter for new streams
         validators,
         accepted_tokens,  # whatever caller wants to accept
         salt=convert(msg.sender, bytes32),  # Ensures unique deployment per caller
```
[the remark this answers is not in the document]
fubuloubu: originally I was thinking that it would make sense to keep it like that and have per-product validators, but I think what I am gonna do is add a secondary `String[32]` "product" variable that also gets concatenated to the salt (and to the `deployments` mapping)

this way, in the StreamFactory Python SDK class I can actually loop through all of the known deployments of the factory (which I want to use the release version identifier as the create2 salt so you can get release by address) and look for the deployments that a particular user has made for a particular product name

contracts/StreamManager.vy
```diff
@@ -17,243 +16,313 @@
...
 @external
-def add_token(token: IERC20):
-    assert msg.sender == self.owner
-    self.token_is_accepted[token] = True
+def accept_control():
+    assert msg.sender == self.new_controller
+    assert block.timestamp - self.new_controller_proposed >= CONTROLLER_ACCEPTANCE_DELAY
+
+    log NewControllerAccepted(self.controller, msg.sender)
+    self.controller = msg.sender
```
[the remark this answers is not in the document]
fubuloubu: should not be consequential, but will do it

also note that the controller can revoke by calling `transfer_control` a second time to themselves (or to the ZERO_ADDRESS)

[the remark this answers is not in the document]
fubuloubu: cc12a80

contracts/StreamManager.vy
```diff
@@ -17,243 +16,313 @@
...
 
 
 @external
-def remove_token(token: IERC20):
-    assert msg.sender == self.owner
-    self.token_is_accepted[token] = False
+def set_capabilities(account: address, capabilities: Ability):
+    if Ability.MODFIY_ACCESS not in self.capabilities[msg.sender]:
```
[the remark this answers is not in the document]
fubuloubu: `flag` is a flag enum, `Ability.MODIFY_ACCESS` is just that one flag set, and yes there is a different semantic here for `not in` that is actually doing `Ability.MODIFY_ACCESS & self.capabilities[msg.sender] == Ability.MODIFY_ACCESS`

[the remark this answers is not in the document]
fubuloubu: So the full logic here is "`self.capabilities` for `msg.sender` has `Ability.MODFIY_ACCESS` flag set"

contracts/StreamManager.vy
fubuloubu: yeah, pricing tokens accurately is indeed a concern w.r.t. their decimals, for example when I priced out our silverback stuff, there was consistently a significant figure in the 7th decimal place that could lead to the price being off ~3% when using USDC (6 decimals) vs. USDT.

However, on the flip side you can deploy different validators that use `token` to do their pricing (maybe all USD-stables is 1 validator, all EUR is a 2nd, etc.), and yeah like you said you can just adjust your pricing slightly to ensure that specific decimals don't bite you too hard in the ass. If al vallidators return 0 then the stream creation will revert.

USDC is annoying lol

sdk/py/apepay/streams.py
fubuloubu: eb9f34a

fubuloubu:
> something weird I noticed, can't tell if it because of Ape or the ApePay library, but if you run the tests w/o contracts being compiled, it will compile them, but then all tests will fail with `'Project' object has no attribute 'StreamManager'`
> 
> Edit: ok this hasn't happened in my sample project, hmmm

yeah I think it might happen if you reference it as a local package via but not when it is bundled since the bundle will copy the file pointed by the symbolic link

I'm not sure if there's a better, more ape-y way to do it, but I would love to find a common pattern here to suggest to any other Ape py SDK deployooooors

contracts/StreamManager.vy
fubuloubu: This goes back to your point about better docs, I think that's the resolution here

contracts/StreamManager.vy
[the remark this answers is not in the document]
mikeshultz: I am not really following this here. Maybe "flag enum" is something I haven't run into. Is it some kind of bit shift packing? 

The Vyper docs just make it sound like an oddly indexed enum.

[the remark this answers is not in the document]
fubuloubu: Inspired by https://docs.python.org/3/library/enum.html#enum.Flag

contracts/StreamManager.vy
fubuloubu: 7c25276

fubuloubu:
> suggestion: document more. Maybe file-level docstrings in the header, or maybe in the README.

docs for `StreamManager` added here: 7c25276

sdk/py/apepay/manager.py
fubuloubu: b8b07d0

fubuloubu:
@mikeshultz don't need a full audit here, but please check for any gotchas you can think of right now before we put this into live testing (on Sepolia first through Platform staging)

contracts/StreamManager.vy
```diff
@@ -1,259 +1,464 @@
...
 
+# Global index of Streams
+num_streams: public(uint256)
+streams: public(HashMap[uint256, Stream])
+
+# Service Provider (has all Capabilities, also beneficiary of funding)
+controller: public(address)
+new_controller: public(address)
+new_controller_proposed: public(uint256)
+CONTROLLER_ACCEPTANCE_DELAY: constant(uint256) = 7 * 24 * 60 * 60  # 1 week
```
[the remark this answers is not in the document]
fubuloubu: No it gets evaluated during compilation

contracts/StreamManager.vy
```diff
@@ -1,259 +1,464 @@
...
 token_is_accepted: public(HashMap[IERC20, bool])
 
+# Global index of Streams
+num_streams: public(uint256)
+streams: public(HashMap[uint256, Stream])
+
+# Service Provider (has all Capabilities, also beneficiary of funding)
+controller: public(address)
+new_controller: public(address)
+new_controller_proposed: public(uint256)
```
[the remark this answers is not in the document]
fubuloubu: Its a timestamp, but I agree the name is suboptimal

contracts/StreamManager.vy
```diff
@@ -1,259 +1,464 @@
...
+    @dev Begin the transfer of the `controller` role to `new_controller`.
+    @notice This action is very dangerous! Can only be performed by the `controller`. After the
+        transition is initiated, the `new_controller` must wait `CONTROLLER_ACCEPTANCE_DELAY`
+        before they are able to accept via `accept_control()`.
+    @param new_controller The address of the proposed new `controller` for this contract.
+    """
+    # NOTE: can revoke transfer at any time calling this method with `self.controller`
+    assert msg.sender == self.controller  # dev: not controller
+
+    log NewControllerProposed(msg.sender, new_controller)
```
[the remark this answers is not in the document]
fubuloubu: Realistically no, because logs can only be read off-chain

But for style purposes I can flip it

contracts/StreamManager.vy
```diff
@@ -1,259 +1,464 @@
...
+
+    self.new_controller = new_controller
+    self.new_controller_proposed = block.timestamp
 
 
 @external
-def add_token(token: IERC20):
-    assert msg.sender == self.owner
-    self.token_is_accepted[token] = True
+def accept_control():
```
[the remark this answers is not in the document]
fubuloubu: The foundational security property is that `controller` is the provider of some service, so if you feel your service would be interrupted by the change then you've 7 days to react to it

But doubtful, this is more of a utilitarian shift for a critical role in case you want to move wallets around. It is so critical though, hence 7 day revocation before acceptance

contracts/StreamManager.vy
```diff
@@ -1,259 +1,464 @@
...
+    @dev Accept `controller` role and responsibilities.
+    @notice This action is very dangerous! Can only be accepted by `new_controller` after waiting
+        for a period of `CONTROLLER_ACCEPTANCE_DELAY`. Once accepted, all Streams claimed in the
+        future will be routed to the new `controller` instead of the previous one. They also will
+        have full and unconditional control over access capabilities for any other address. Note
+        that `controller` can prevent this action at any time up to `CONTROLLER_ACCEPTANCE_DELAY`
+        by executing `transfer_control()` with themselves (or a different address) as the proposed.
+    """
+    assert msg.sender == self.new_controller  # dev: not proposed controller
+    assert block.timestamp - self.new_controller_proposed >= CONTROLLER_ACCEPTANCE_DELAY
```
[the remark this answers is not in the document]
fubuloubu: My style has been to use dev messages for uncommon things, or things younahpuld know about, and use revert messages for things that are likely to happen

contracts/StreamManager.vy
```diff
@@ -1,259 +1,464 @@
...
+        account that has the `MODFIY_ACCESS` capability. Please note that any capabilities granted
+        go into effect immediately. Any capability can be reverted at any time, but please note
+        if granting the `MODFIY_ACCESS` capability to any account, they will have the capability to
+        themselves grant that (and any other role) to any other account, presenting a potential
+        DDoS risk to the `controller`, and of course other dangerous capabilities that could lead
+        to security risks of other natures (such as malcious tokens, validators which cause DDoS).
+    @param account The address of the account to grant or revoke `capabilities` to.
+    @param capabilities The new set of abilities that `account` should have access to.
+    """
+    if Ability.MODFIY_ACCESS not in self.capabilities[msg.sender]:
```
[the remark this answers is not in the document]
fubuloubu: Potentially, but also a long statement

contracts/StreamManager.vy
```diff
@@ -1,259 +1,464 @@
...
+    @dev Assign the set of validators that should be executed on Stream creation.
+    @notice This can only be called by the controller or someone with the MODFIY_VALIDATORS
+        capability. It is suggested to ensure that the array contains all unique entries, lest
+        unpredictable or actively harmful conditions may happen.
+    @param validators The array of validators to assign for this contract.
+    """
+    if Ability.MODFIY_VALIDATORS not in self.capabilities[msg.sender]:
+        assert msg.sender == self.controller  # dev: insufficient capability
+
+    # TODO: Ensure uniqueness using Set
```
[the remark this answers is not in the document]
fubuloubu: Vyper doesn't have Set yet

This not critical, I use `set` in the Python SDK, but if you scrrw it up it won't break anything, it just may charge your users double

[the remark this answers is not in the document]
fubuloubu: I was considering just doing the `O(n**2)` brute force check though

contracts/StreamManager.vy
```diff
@@ -1,259 +1,464 @@
...
+        very small values can cause problems since this contracts works with streaming rates of
+        token/second. For example, tokens like USDC can cause problems if the value of the product
+        is lower than $30/month (vs. USDT which has 18 decimals).
+    @param token An ERC20-compatible token to accept or reject.
+    @param is_accepted A boolean value that controls whehter `token` is accepted or rejected.
+    """
+    if Ability.MODFIY_TOKENS not in self.capabilities[msg.sender]:
+        assert msg.sender == self.controller  # dev: insufficient capability
+
+    self.token_is_accepted[token] = is_accepted
```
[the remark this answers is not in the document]
fubuloubu: That is correct, resetting a boolean mapping to its default value is exactly the same as setting it false

contracts/StreamManager.vy
```diff
@@ -1,259 +1,464 @@
...
+        # NOTE: Validator either raises or returns a funding rate to add to the total
+        amount_per_second += extcall validator.validate(msg.sender, token, amount, products)
+
+    # Ensure stream life parameters are acceptable to caller
+    # NOTE: div/0 if `amount_per_second` is 0, signaling no supported products found
+    stream_life: uint256 = amount // amount_per_second  # dev: no valid products detected
+    assert min_stream_life <= stream_life  # dev: stream too expensive
+
+    # Create stream data structure and start streaming
+    stream_id: uint256 = self.num_streams
```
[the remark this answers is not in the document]
fubuloubu: Yes this would function fine because each transaction is atomic and the first updates the value prior to the second one executing at all

contracts/StreamManager.vy
```diff
@@ -1,259 +1,464 @@
...
+        msg.sender, self, amount, default_return_value=True
     )
 
-    stream_id: uint256 = self.num_streams[msg.sender]
-    self.streams[msg.sender][stream_id] = Stream({
+    # Check all validators for any unacceptable or incorrect stream parameters
+    amount_per_second: uint256 = 0
+    for validator: Validator in self.validators:
+        # NOTE: Validator either raises or returns a funding rate to add to the total
+        amount_per_second += extcall validator.validate(msg.sender, token, amount, products)
```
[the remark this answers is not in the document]
fubuloubu: I think of the set of validators as the way to check and validate stream parameters to compute the subtotal cost of a stream. There may be different product codes handled by different validators

How do you think I can make this clearer?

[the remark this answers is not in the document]
antazoey: Each validators gets the full list of products, maybe that is why I am confused. If it were the case of like mapping products to validators, it might be less confusing

contracts/StreamManager.vy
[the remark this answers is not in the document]
mikeshultz: Never ran into this before.  Thanks.

contracts/StreamManager.vy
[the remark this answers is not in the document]
mikeshultz: I kind of agree validators could be done a bit better, maybe. And they're conceptually a little confusing. Just *adding* to the price doesn't seem very useful.

When I first thought of validators, I was thinking about things that can approve/reject/modify the terms of the stream. For instance, it might make sense to write one up that sets discounts for certain users/times. And you might have another you might want to check an exchange rate with uniswap before returning a price.

For these concepts to work, you'd need to be able to modify the price, not just add to it.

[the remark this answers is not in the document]
fubuloubu: Actually, this gave me some ideas that I think can eliminate a few footguns and overall make it simpler. Will refactor and push it up

fubuloubu:
note: coverage/gas reporting seems to be failing erronously

contracts/StreamManager.vy
```diff
@@ -1,259 +1,532 @@
...
+    assert extcall token.transferFrom(
+        msg.sender, self, amount, default_return_value=True
+    )
+    assert block.timestamp < self.streams[stream_id].expires_at
 
-    funded_amount: uint256 = self.streams[creator][stream_id].funded_amount
-    amount_locked: uint256 = funded_amount - self._amount_unlocked(creator, stream_id)
-    assert amount_locked > 0  # NOTE: reverts if stream doesn't exist, or already cancelled
-    self.streams[creator][stream_id].funded_amount = funded_amount - amount_locked
+    # Make sure stream claims are up-to-date (ensures that stream rate math is correct)
```
fubuloubu, line 441: Make a note that this is an important security property because if the funding rate changes then the math for claiming will no longer be accurate.

contracts/StreamManager.vy
```diff
@@ -1,259 +1,464 @@
...
+
+# Service Provider (has all Capabilities, also beneficiary of funding)
+controller: public(address)
+new_controller: public(address)
+new_controller_proposed: public(uint256)
+CONTROLLER_ACCEPTANCE_DELAY: constant(uint256) = 7 * 24 * 60 * 60  # 1 week
+
+event NewControllerProposed:
+    old: indexed(address)
+    new: indexed(address)
```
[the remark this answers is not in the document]
fubuloubu: did this, and also for the stream owner update event, and also made all the stream events use `stream_id` instead of just `id` (it's clearer)

5d17c1b
8cb68c9
dbdd779

contracts/StreamManager.vy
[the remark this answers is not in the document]
fubuloubu: 9262ad4

contracts/StreamManager.vy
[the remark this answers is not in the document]
fubuloubu: cd9134c

contracts/StreamManager.vy
fubuloubu:
```suggestion
    # NOTE: Stream claims must be up-to-date to ensure that math is correct in `_amount_claimable()`
    #       This is because the stream rate may change in `_compute_stream_life()`
```

contracts/Validator.vyi
```diff
@@ -1,13 +1,43 @@
...
+    creation, funding or migration events in order to block certain behaviors or modify the stream
+    time to a specific limit, as in the case of a pre-release or sunsetting product.
+"""
 from ethereum.ercs import IERC20
 
-MAX_REASON_SIZE: constant(uint16) = 1024
-
 
 @external
 def validate(
```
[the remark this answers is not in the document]
fubuloubu: hopefully this helps?
[1f46007](https://github.com/ApeWorX/ApePay/pull/112/commits/1f460073e0252b1d3bc4af1dbb0b98b8172e79f6)

contracts/StreamManager.vy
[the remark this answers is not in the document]
fubuloubu: implemented in [9674830](https://github.com/ApeWorX/ApePay/pull/112/commits/9674830d7044aeecf499a1dd10946f2dfdcedb32) and further commits, please review

fubuloubu:
Deployed a preliminary cut here (using salt `"ApePay 0.5.0rc1"`):
blueprint - https://sepolia.etherscan.io/address/0xA4e75aef1362A5c037a43709E9cdB1649D9F1303
factory - https://sepolia.etherscan.io/address/0xacf6369f76f04ec96ce881423e858805adbd2bd5
manager - https://sepolia.etherscan.io/address/0x7B25175142886e34543247f4583fe034eafDee32

tests/conftest.py
```diff
@@ -21,53 +29,92 @@ def create_token(deployer):
...
+        return stream_manager.create(token, amount, products, sender=(sender or payer), **txn_args)
+
+    return create_stream
+
+
 @pytest.fixture(scope="session")
-def stream(stream_manager, token, payer):
-    return stream_manager.create(token, 1000, sender=payer)
+def stream(chain, create_stream):
+    # TODO: Remove when https://github.com/ApeWorX/ape/pull/2277 merges
```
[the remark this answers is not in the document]
antazoey: sorry i see you made some changes here since i had figured this out.
but moving the isolation to session scoped stream fixture doesnt change much.

the interlocking of those parametrized tests causes things to run in that order.
create stream will not be 0 becase cancel stream would have triggered the a session scoped fixure

but making the session scoped fixture wrapped in isolation, it essentially is undoing itself much like a function scoped fixture

contracts/StreamManager.vy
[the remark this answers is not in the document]
mikeshultz: I'm not sure how this changes much on the validator front.  Now instead of validators only adding to the stream cost, they can only add to the stream duration?

That's fine with me, if that's what you're going for. I'm just not really sure how this makes validators more useful or understandable.

contracts/StreamManager.vy
```diff
@@ -1,259 +1,539 @@
...
 
 
-event StreamCancelled:
-    creator: indexed(address)
+event StreamClaimed:
     stream_id: indexed(uint256)
-    amount_locked: uint256
-    reason: Bytes[MAX_REASON_SIZE]
+    claimer: indexed(address)
+    is_expires: indexed(bool)
```
mikeshultz, line 106: issue: I'm assuming this should be `is_expired`?

fubuloubu:
```suggestion
    is_expired: indexed(bool)
```

tests/conftest.py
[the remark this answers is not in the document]
fubuloubu: yes, I think this is a good point, that this fixture should not be `session`

sdk/py/apepay/manager.py
antazoey: typically you import it `from ape.logging import logger`

antazoey: Actually another under-rated approach is subclass `ApeCliContextObject` especially if this class is designed for CLI usage only
And you can use it in your CLIs like:

```python
@ape_cli_context(obj_type=Runner)
```

sdk/py/apepay/manager.py
```diff
@@ -87,132 +91,131 @@ def _parse_validator(self) -> Callable[[_ValidatorItem], Validator]:
...
-    def remove_token(self) -> ContractTransactionHandler:
-        return self.contract.remove_token
+    def remove_token(self, token: AddressType, **txn_kwargs) -> ReceiptAPI:
+        return self.contract.set_token_accepted(token, False, **txn_kwargs)
 
-    @property
-    def is_accepted(self) -> ContractCallHandler:
-        return self.contract.token_is_accepted
+    def is_accepted(self, token: AddressType) -> bool:
+        return self.contract.token_is_accepted(token)
```
antazoey, line 138: We could potentially use the ape-extras mechanism and feed into `.contract` so all pass throughs are automatic.

contracts/StreamManager.vy
```diff
@@ -1,259 +1,539 @@
...
+
 
 @external
-def set_validators(validators: DynArray[Validator, MAX_VALIDATORS]):
-    assert msg.sender == self.owner
-    self.validators = validators
+def transfer_control(new_controller: address):
+    """
+    @dev Begin the transfer of the `controller` role to `new_controller`.
+    @notice This action is very dangerous! Can only be performed by the `controller`. After the
```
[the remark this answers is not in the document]
fubuloubu: Yes, they can transfer control back to themselves, and just never accept it

sdk/py/apepay/manager.py
fubuloubu: How does that work again?

contracts/Validator.json
```diff
@@ -4,11 +4,11 @@
     "name": "validate",
     "stateMutability": "nonpayable",
     "inputs": [
-      { "name": "creator", "type": "address" },
+      { "name": "funder", "type": "address" },
```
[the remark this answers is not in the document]
fubuloubu: yeah this may just not be supported right by the tooling (until it's fixed in 0.4.1)

contracts/StreamFactory.vy
```diff
@@ -16,24 +16,35 @@ BLUEPRINT: public(immutable(address))
...
         code_offset=3,
     )
     self.deployments[msg.sender] = deployment
 
+    log ManagerCreated(msg.sender, deployment, accepted_tokens, validators)
```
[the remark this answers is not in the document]
fubuloubu: d989662

PR #115, opened 2024-10-24
fix: import BaseInterfaceModel from correct location

antazoey:
### What I did

It is only incidental that this even works, all the basemodel stuff is defined in `ape.utils.basemodel` and has been since its existence. Something in `ape.types.__init__` is using `BaseInterfaceModel` which is why it was working.

It is not even listed in the `__all__` https://github.com/ApeWorX/ape/blob/v0.8.17/src/ape/types/__init__.py#L573 so I am surprised you don't at least get static import warnings.

### How I did it

Import from the correct spot.

### Checklist

- [ ] All changes are completed
- [ ] New test cases have been added
- [ ] Documentation has been updated

### DIFF
--- a/sdk/py/apepay/factory.py
+++ b/sdk/py/apepay/factory.py
@@ -1,8 +1,8 @@
 from typing import Any
 
 from ape.contracts import ContractInstance
-from ape.types import AddressType, BaseInterfaceModel
-from ape.utils import ZERO_ADDRESS
+from ape.types import AddressType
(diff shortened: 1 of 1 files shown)


fubuloubu (approved):

fubuloubu:
Is this critical to release?

antazoey:
> Is this critical to release?

No, and I am even thinking on exporting the base-model stuff from the root of `ape.types` now because of this..
Thinking of making that the official way to do it in 0.9.

